PR #98, opened 2025-02-26
Refactor `MaybeTaken` to use `Option`

laurmaedje:
Small refactor to make the implementation of `MaybeTaken` a bit more straightforward and the error messages in the wrong case a bit better. I first noticed that the `RefCell` could just be a `Cell`, but then it turned out that using an `Option` is even simpler and also relaxes the technically unnecessary `Default` bound. I also added `#[track_caller]` so that the error is shown at the wrong usage if it goes wrong.


LaurenzV:
Uhmm good point, no idea anymore how I ended up with that. :D Thanks!

PR #100, opened 2025-02-26
Refactor chunk remapping

laurmaedje:
I found the chunk remapping macros quite hard to read, so I figured I'd attempt a small refactor here. The refactor converts the macros and local variables into a `Visit` trait that visits all chunks in a type. The field order is now defined only once and we have two for loops for the remapping and writing. This also fixes the bug that the PDF file size guess was always zero because `chunks_len` was read before remapping even occurs.

### DIFF
--- a/crates/krilla/src/chunk_container.rs
+++ b/crates/krilla/src/chunk_container.rs
@@ -12,32 +12,6 @@ use crate::util::{hash_base64, Deferred};
 use crate::validation::ValidationError;
 use crate::version::PdfVersion;
 
-trait ResExt {
-    fn res(&self) -> KrillaResult<&Chunk>;
-}
-
-trait WaitExt {
-    fn wait(&self) -> &Chunk;
-}
-
-impl ResExt for Chunk {
-    fn res(&self) -> KrillaResult<&Chunk> {
-        Ok(self)
-    }
-}
-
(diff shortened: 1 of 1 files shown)


crates/krilla/src/chunk_container.rs
```diff
@@ -131,44 +82,10 @@ impl ChunkContainer {
...
-            &self.page_label_tree, &self.destination_profiles,
-        &mut self.struct_tree_root);
-        write_fields!(remapper, &mut pdf; &self.struct_elements, &self.page_labels,
-            &self.annotations, &self.fonts, &self.color_spaces, &self.icc_profiles, &self.destinations,
-            &self.ext_g_states, &self.masks, &self.x_objects,
-            &self.shading_functions, &self.patterns, &self.pages, &self.images
-        );
+        // Write the chunks in all the fields.
+        self.visit(&mut |chunk| {
+            chunk.renumber_into(&mut pdf, |old| remapper[&old]);
```
LaurenzV, line 87: Wow, I never knew that you could access items in a HashMap with `[]`! 😅

laurmaedje: Only works for immutable access!

LaurenzV:
I think this does look nicer, thanks!

PR #103, opened 2025-02-27
Add support for image interpolation

LaurenzV:


crates/krilla/src/validation/PDF_A1.md
```diff
@@ -80,7 +80,9 @@ in case CMYK is used but no profile was provided, export fails. 🟢
 
 6.2.3.4: krilla does not support DeviceN/Separation color spaces. 🔵
 
-6.2.4: krilla does not use the `Alternates`/`Interpolate`/`Intent` keys for images. 🔵
+6.2.4: 
+- krilla does not use the `Alternates`/`Intent` keys for images. 🔵
+- krilla does check whether the `Interpolate` is used. 🟢
```
laurmaedje, line 85:
```suggestion
- krilla does check whether the `Interpolate` key is used. 🟢
```

crates/krilla/src/validation/PDF_A2.md
```diff
@@ -86,7 +86,9 @@ in case CMYK is used but no profile was provided, export fails.
 
 6.2.7: krilla is not a reader. 🔵
 
-6.2.8.1: krilla does not use the `Alternates`/`Interpolate`/`Intent` keys for images. 🔵
+6.2.8.1: 
+- krilla does not use the `Alternates`/`Intent` keys for images. 🔵
+- krilla does check whether the `Interpolate` is used. 🟢
```
laurmaedje, line 91:
```suggestion
- krilla does check whether the `Interpolate` key is used. 🟢
```

crates/krilla/src/svg/image.rs
```diff
@@ -25,19 +25,19 @@ pub(crate) fn render(
 
     match image.kind() {
         ImageKind::JPEG(d) => {
-            let image = Image::from_jpeg(d.clone())?;
+            let image = Image::from_jpeg(d.clone(), false)?;
```
laurmaedje, line 28: Technically, I wonder whether it would make sense to recursively apply `/Interpolate` to raster images in SVG if the SVG itself has smooth scaling in Typst, but not like it happens currently.

LaurenzV: We could do that, though it might be a bit unfortunate if someone wants to export to PDF/A and they can't because their SVG happens to use smooth scaling for images, even though it's not really necessary.

laurmaedje: > happens to use smooth scaling for images

If it's only about whether `set image(scaling: "smooth")` is set or not, they would have control over it, wouldn't it?

But anyway, it's not too important for now.

laurmaedje (approved):

crates/krilla/src/svg/image.rs
LaurenzV: Ahh, I thought you were talking about the SVG attribute which allows to define the interpolation used for images.

PR #105, opened 2025-02-27
Add support for file embedding

LaurenzV:


crates/krilla/src/object/embed.rs
```diff
@@ -0,0 +1,181 @@
...
+    MissingDate,
+}
+
+/// An embedded file.
+#[derive(Debug, Clone, Hash)]
+pub struct EmbeddedFile {
+    /// The name of the embedded file.
+    pub path: String,
+    /// The mime type of the embedded file.
+    pub mime_type: String,
```
laurmaedje, line 28: The MIME type is optional unless it's PDF/A-3.

crates/krilla/src/object/embed.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// A description of the embedded file.
+    pub description: Option<String>,
+    /// The association kind of the embedded file.
+    pub association_kind: AssociationKind,
+    /// The raw data of the embedded file.
+    pub data: Arc<Vec<u8>>,
```
laurmaedje, line 34: This is more of a general comment: It would be nice if krilla would take a more generic type like `Arc<dyn AsRef<[u8]> + Send + Sync>` or some custom container like that, so that one doesn't have to copy the data if the source representation differs a bit.

crates/krilla/src/object/embed.rs
```diff
@@ -0,0 +1,181 @@
...
+        params.size(self.data.len() as i32);
+
+        if let Some(date_time) = sc
+            .metadata()
+            .and_then(|m| m.modification_date.or_else(|| m.creation_date))
+        {
+            let date = pdf_date(date_time);
+            params.modification_date(date);
+        } else {
+        }
```
laurmaedje, line 70: what is happening here?

LaurenzV: Just haven't implemented it yet, will be added once I add the PDF/A3 checks

PR #135, opened 2025-03-11
Tidy imports

laurmaedje:
I know it's bad practice to start contributing to a project and dumping a formatting PR on the maintainer, sorry 😅

I just saw that it's bascially std/extern/crate formatting style, but not consistently.

### DIFF
--- a/crates/krilla/src/configure/validate.rs
+++ b/crates/krilla/src/configure/validate.rs
@@ -20,15 +20,17 @@
 //!
 //! [`Configuration`]: crate::configure::Configuration
 
+use std::fmt::Debug;
+
+use pdf_writer::types::OutputIntentSubtype;
+use pdf_writer::Finish;
+use skrifa::GlyphId;
+use xmp_writer::XmpWriter;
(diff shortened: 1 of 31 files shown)


LaurenzV:
Yeah at some point I reformatted everything to be in that order, but over time I slacked off. 😅

LaurenzV:
Is there a way to enforce that automatically? Or do you just manually rearrange?

LaurenzV:
Well that's embarassing. 😅 I think the first time I spent like 15 minutes reformatting the imports manually this way... Good to know.

LaurenzV:
For future reference: `cargo fmt -- --config group_imports=StdExternalCrate`

laurmaedje:
> Well that's embarassing. 😅 I think the first time I spent like 15 minutes reformatting the imports manually this way... Good to know.

I did it manually as well :p

laurmaedje:
I think I tried the cargo thing, but disliked how it grouped `pub use` with normal `use`.

PR #139, opened 2025-03-12
Completely restructure crate setup

LaurenzV:
An attempt to completely restructure the crate, including splitting off testing and perhaps extracting SVG handling into a custom crate. Will probably take a while to finish.


LaurenzV:
Crate is still in a rough state, but a good chunk of the refactoring is done, so I'll merge this for now to have a "checkpoint" in main.

PR #169, opened 2025-03-24
Replace `cmap_inner!` macro by generic function

laurmaedje:
Not that it matters much, but a bit unfortunate that it had to be a macro rather than a generic function. I tried to rewrite and noticed that `pdf-writer` doesn't properly export its `GlyphId` trait. It's still possible to still write it generically, but it requires introduction of a helper trait. That trait could be removed if the missing reexport is added in `pdf-writer`. ~~Marking as draft because fixing the reexport and bumping pdf-writer would be easy enough.~~

### DIFF
--- a/crates/krilla/src/text/cid.rs
+++ b/crates/krilla/src/text/cid.rs
@@ -30,59 +30,62 @@ pub(crate) const SYSTEM_INFO: SystemInfo = SystemInfo {
 
 pub(crate) type Cid = u16;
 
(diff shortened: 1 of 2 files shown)


laurmaedje:
I made the trait public in pdf-writer, but krilla currently still depends on the unmerged PR https://github.com/typst/pdf-writer/pull/47. I left another review there to move it forward.

PR #171, opened 2025-03-26
Minor fixes

laurmaedje:
Primarily to silence clippy


LaurenzV:
That's weird, shouldn't it be detected in CI?

laurmaedje:
Ah, I was wondering too. Turns out I had accidentally set directory overrides that made it use the nightly toolchain in the krilla repo.

LaurenzV:
I mean, it doesn't hurt.^^

PR #180, opened 2025-03-27
Use `flate2` with `zlib-rs` backend for deflating

LaurenzV:
It's much faster.


LaurenzV:
I'll go ahead with the merge so I can keep testing, if you have any objections we can still revert it later on. ^^

laurmaedje:
There is also the consideration that png still depends on miniz_oxide so we have both now, but that small extra dependency burden is not that big of a deal if this provides meaningful speedups. Also, regarding binary size, one could hope that miniz_oxide's encoding and zlib-rs's decoding part is eliminated in dead code elimination. No idea whether it happens in practice.

PR #182, opened 2025-03-28
Bump subsetter

LaurenzV:
Unfortunately requires regenerating all tests becuase the subset tags are different now.

### DIFF
--- a/crates/krilla/src/text/cid.rs
+++ b/crates/krilla/src/text/cid.rs
@@ -133,8 +133,7 @@ impl CIDFont {
 
     #[inline]
     pub(crate) fn get_cid(&self, glyph_id: GlyphId) -> Option<u16> {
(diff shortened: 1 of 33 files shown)


crates/krilla/src/text/cid.rs
```diff
@@ -134,7 +134,7 @@ impl CIDFont {
     #[inline]
     pub(crate) fn get_cid(&self, glyph_id: GlyphId) -> Option<u16> {
         self.glyph_remapper
-            .get(u16::try_from(glyph_id.to_u32()).unwrap())
+            .get(glyph_id.to_u32() as u16)
```
LaurenzV, line 137: I think I also measured a speedup there, and since glyph IDs with > u16 don't exist yet, this isn't a problem

PR #189, opened 2025-04-01
Make font writing deterministic

LaurenzV:


laurmaedje (commented):
I think this is fine and we can merge it, though I would probably have just retained the input order instead of sorting by hash.

The sorting by hash only ensures stability if the hash is stable across runs. For hashes of just strings or bytes that's trivially the case, but for more complex data structures, it adds an extra unnecessary requirement on the hashing that isn't strictly needed. In Typst, for example, we sometimes hash pointers to 'static items and the hashes are not stable across two CLI invocations because of ASLR (address space layout randomization). But Typst doesn't rely on the exact values of those hashes anywhere in the output, so it's still fine. 

Of course, what Typst does isn't all that relevant here, just trying to explain where I'm coming from.

PR #200, opened 2025-04-28
Add PageLayout to Metadata

maltevonehren:
This MR exposes the /PageLayout property of the Document Catalog through the document metadata and adds it to the metadata_full snapshot tests.

I added copied the struct definition for `PageLayout` from `pdf-writer`, analogously to TextDirection (the other option would be to `pub use` the type from `pdf-writer`)

### DIFF
--- a/crates/krilla/src/chunk_container.rs
+++ b/crates/krilla/src/chunk_container.rs
@@ -6,6 +6,7 @@ use xmp_writer::{RenditionClass, XmpWriter};
 use crate::configure::{PdfVersion, ValidationError};
 use crate::error::KrillaResult;
 use crate::interchange::metadata::Metadata;
+use crate::metadata::PageLayout;
 use crate::serialize::SerializeContext;
 use crate::util::{hash_base64, Deferred};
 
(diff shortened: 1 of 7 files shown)


crates/krilla/src/chunk_container.rs
```diff
@@ -220,6 +220,11 @@ impl ChunkContainer {
                 }
             }
 
+            let page_layout = self.metadata.as_ref().and_then(|m| m.page_layout);
+            if let Some(layout) = page_layout {
+                catalog.page_layout(layout.to_pdf());
+            }
+
```
LaurenzV, line 227: For the two page layouts that require PDF 1.5, we need a check whether the version is high enough, and if not we should not write the attribute.

LaurenzV (commented):
Looks good, just one comment.

LaurenzV:
> (the other option would be to pub use the type from pdf-writer)

We are doing this on purpose to avoid re-exporting pdf-writer and thus having to make breaking releases if we bump the dependency. So the way you did it now is fine!

LaurenzV:
Thanks!

PR #202, opened 2025-05-03
Add draw_svg_with_position method to specify the placement of the SVG

koba-ninkigumi:
Add draw_svg_with_position method to specify the placement of the SVG

### DIFF
--- a/crates/krilla-svg/src/lib.rs
+++ b/crates/krilla-svg/src/lib.rs
@@ -15,7 +15,7 @@ use std::sync::Arc;
 
 use fontdb::Database;
 use krilla::color::rgb;
-use krilla::geom::{Rect, Size, Transform};
+use krilla::geom::{Point, Rect, Size, Transform};
 use krilla::paint::FillRule;
 use krilla::surface::Surface;
 use krilla::text::Font;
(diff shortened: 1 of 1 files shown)


LaurenzV:
I'm not sure I understand the need for this, can't you just push a translation transform before drawing the SVG?

koba-ninkigumi:
First, from the perspective of consistency:

If we adopt your logic, we should remove the argument for specifying the position in the draw_text method.

Second, from the perspective of absolute positioning:

If we use the translation transform as you suggested, when placing multiple SVGs with absolute positions on the same page, we would need to first move to the desired position using the translation transform, place the SVG, then reset the position back to (0.0, 0.0), move to the next specified position, place the next SVG, and again reset the position to (0.0, 0.0).

Third, considering the reverse perspective:

What if the original draw_svg method was one that allowed specifying positions, similar to how draw_text works? In that case, if we were to have a pull request suggesting changing the draw_svg method to one where the position is fixed at (0.0, 0.0) because of the existing translation transform feature, how would we judge that change?

LaurenzV:
> If we adopt your logic, we should remove the argument for specifying the position in the draw_text method.

While I can see your point about consistency, I do think those are a bit different because positioning text in different locations is much more common (you might have a few dozens of images, but you will have thousands or tens of thousands of line of text positioned in different places in big documents), so I think it makes sense to have it. In addition to that, using the position parameter for text allows you to shift text with gradient/pattern fills without transforming the underlying paint, which is also pretty important, but not relevant for SVG images. And thirdly, for consistency we would have to do the same for `draw_image` as well, and I think I would rather not pollute the API more. :)

> If we use the translation transform as you suggested, when placing multiple SVGs with absolute positions on the same page, we would need to first move to the desired position using the translation transform, place the SVG, then reset the position back to (0.0, 0.0), move to the next specified position, place the next SVG, and again reset the position to (0.0, 0.0).

That's true, but it's only two lines of additional code. You just do
```rs
surface.push_transform(...)
surface.draw_svg(...)
surface.pop_transform)
```

And I don't think that's too bad. So, thanks a lot for the PR, but I think my preference here is to keep the current API. If you have trouble getting the positioning to work correctly, feel free to let me know and I can provide you with an example. :)

PR #203, opened 2025-05-08
Add option to automatically chose whether to compress embedded files

saecki:
Continuing from https://github.com/typst/typst/issues/5762, is this what you had in mind? If  yes I would go ahead, update the tests and make this ready for review.


crates/krilla/src/interchange/embed.rs
```diff
@@ -41,8 +41,9 @@ pub struct EmbeddedFile {
     /// The raw data of the embedded file.
     pub data: Data,
     /// Whether the embedded file should be compressed (recommended to turn off if the
-    /// original file already has compression).
-    pub compress: bool,
+    /// original file already has compression). If [`None`] the file will be compressed
+    /// and if no sufficient improvements were noted, the uncompressed file is embedded.
```
LaurenzV, line 45: I would just say `If None, krilla will use its own logic for determining whether to compress the file or not`.

LaurenzV:
Thanks! There is the problem now that the snapshot is not ASCII compatible anymore, but you don't need to figure this out now and I'll look into it in the near future! I just need to figure out why I broke CI, then I'll merge.

PR #213, opened 2025-06-23
Bump `skrifa`

LaurenzV:


LaurenzV:
Okay, well, the reason the one image is different is that the current skrifa code uses a different logic for determining the y-bearing  of sbix glyphs (based on Skia). However, from testing locally this doesn't seem to affect Apple Color Emoji (which is the main `sbix` font we care about), so this should be fine.

PR #214, opened 2025-06-26
Support headings tags `Hn` with `n > 7`

saecki:
The PDF/UA-1 spec `7.4.3 Additional headings` suggests defining `Hn` for level 7 and upwards.
This PR implements this by dynamically registering any used headings unless PDF 2.0 is used, which natively supports these heading levels.


saecki:
Seems like a bunch of new clippy warnings dropped in rust `1.88.0`.

LaurenzV:
If you rebase onto main it should work now.

LaurenzV:
Thank you!

PR #218, opened 2025-07-10
Fix page index shadowing number index in parent tree

saecki:
This fixes a bug where the wrong integer was written into the `ParentTree` map.

It's marked as a draft, because I've rebased onto #217, since both of them touch the same snapshot test.


LaurenzV:
image
See here, it looks like the model checker isn't happy with that PDF. Maybe you could check the differences and see whether it's really correct?

saecki:
Hmm, you're right :thinking:

saecki:
I'll look into this more. Don't feel forced to respond to the issues and PRs right away.

PR #220, opened 2025-07-12
Fix annotation struct parent pointing to itself

saecki:
The annotation's struct parent ref is stored alongside the annotation ref when serializing the tag tree and then read when writing the parent tree in the struct tree root.

From the PDF 1.7 spec, I could also read it the way it was implemented before:
> (Required if the annotation is a structural content item; PDF 1.3) The
integer key of the annotation’s entry in the structural parent tree (see
14.7.4.4, “Finding Structure Elements from Content Items”).

But looking at the tags of: https://pdfa.org/resource/tagged-pdf-best-practice-guide-syntax/
and validating the file in VeraPDF, this seems to be the intended way.


saecki:
Ok, I've read a bit further and found this in the spec which makes it quite clear :sweat_smile:
> For an object identified as a content item by means of an object reference (see 14.7.5.3, "PDF
objects as content items"), the value shall be an indirect reference to the parent structure element.

PR #226, opened 2025-07-17
Automatically generate slightly expanded bounding box for link annotations with quadpoints

saecki:
This adds a new `Quadrilateral` type that documents the order in which points should appear.
Instead of the previous `new` function there are now two different functions to construct a `LinkAnnotation`.
One in which just a `Rect` is provided and the other where a non-empty list of `Quadrilateral`s is passed.
The bounding `rect` is computed automatically for the latter.

The choice of `EPSILON` is a little bit random, so feel free to suggest a different number :)


saecki:
Maybe `new` and `with_quadpoints`?

LaurenzV:
Let's do `new_with_quad_points`?

PR #229, opened 2025-07-18
Actually serialize table summary

saecki:
I don't know how I've missed that before, but the table summary currently isn't being serialized.


crates/krilla/src/interchange/tagging.rs
```diff
@@ -901,15 +901,27 @@ impl TagGroup {
             sc.register_validation_error(ValidationError::MissingHeadingTitle);
         }
 
-        match self.tag.kind {
-            TagKind::L(ln) => {
+        match &self.tag.kind {
+            &TagKind::L(ln) => {
```
LaurenzV, line 905: I think that `&` can be removed no?

LaurenzV (approved):

PR #234, opened 2025-07-24
Add layout attributes

reknih:
This PR adds the following to Krilla:

- The structure types `Div` and `Span`, useful for attaching attributes to
- All[^1] standard layout attributes from PDF 1.7[^2] and below

I encountered two problems with the API of `pdf-writer` when creating this PR:

- The `LineHeight` struct is private, so the `.line_height()` method cannot be used
- `Padding`, `BorderColor`, `BorderStyle`, `BorderThickness`, `TBorderStyle`, and `TPadding` allow two forms: One with a value for each side of the container (an array) and a shorthand to set all sides to the same value. `pdf-writer` only offers one of the two modes for each in its typed API and, embarassingly, which one is not consistent

I have implemented workarounds for both problems.

Let me know if you'd like any changes and whether you'd like me to change `pdf-writer` first or if you're fine with landing this.

[^1]:
    The `RubyAlign` and `RubyPosition` attributes have been omitted because the `Ruby` and `Warichu` structure types they apply to are currently not available in Krilla.
[^2]:
    PDF 2.0 has introduced one new layout attribute, `TextPosition` for marking that text should be in sub- or superscript. `pdf-writer` does not currently support this attribute, let me know if I should write it using the manual API.


LaurenzV:
Using workarounds for `pdf-writer` issue seems fine to me, but we should add TODO comments to track them.

Other than that, without having validated all of the attributes myself, it looks good to me, but I definitely would like to see 1-2 snapshot tests that make use of that new functionality to ensure it actually works. :)

(Another thing I noticed is that maybe there is some way to remove some of the duplication in `generate.toml`, but that's not for this PR to figure out.)

PR #241, opened 2025-08-02
Recompute the global bbox of font

LaurenzV:
@laurmaedje What do you think as this for a fix of https://github.com/typst/typst/issues/6671? The ideal thing to do would obivously be to recompute the bbox in the subsetter, but this requires outlining capabilities which we don't have.

Before:
image

After:
image

I've also rechecked some of the other PDF tests that were regenerated in different viewers and didn't notice anything off.

However, the bbox does look a bit tight this way, for example:

Before:
image

After:
image

Not sure if we should maybe add a bit of padding?


laurmaedje:
Could work though I wonder what other software does. 

I also wonder how expensive this is, especially since it seems to draw the full path into memory and then compute it bounds rather than the ttf-parser routine which just gives the bbox when drawing into a dummy outline builder.

PR #248, opened 2025-08-09
chore: update hayro

magistau:
needed for typst/typst#6664

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -514,6 +514,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "euclid"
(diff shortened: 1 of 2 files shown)


LaurenzV:
Have you actually tested whether it works with that patch? I think there is another instance of include_bytes I still need to fix, though it’s only in a conditionally compiled code that is not activated by default.

magistau:
I have not tested it throughoutly (mostly because I don't even have a deep understanding of krilla), but running `nix build` on typst/typst#6741 (which uses the updated hayro and krilla with this patch) works fine.

PR #254, opened 2025-08-15
Remove dead code

laurmaedje:


LaurenzV:
Is that kind of dead code not detected by the Rust compiler? IIRC we do treat warnings as errors in CI. 🤔

PR #260, opened 2025-09-10
Write `/ReversedChars` marked content for RTL text

laurmaedje:
This ensures that drawn glyphs that have monotonically decreasing ranges (indicating RTL text) are surrounded with `/ReversedChars` marked content as per PDF 1.7 § 14.8.2.3.3. I couldn't detect any difference with this in readers (Acrobat managed to copy-paste Arabic text perfectly fine before), but the spec wants it and it makes sense that the reader should somehow know about the logical glyph order.

We could also introduce a new argument for this, but I figured this was the simplest way to detect the situation and it is in line with what [Skia](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/pdf/SkClusterator.cpp;l=15-27;drc=bea3840c4307aea5cb9a691b4c1f48db1882d9e9) does for this.

The spec also says

> The show strings may have a SPACE (U+0020) character at the beginning or end to indicate a word break (see
14.8.2.5, “Identifying Word Breaks”) but shall not contain interior SPACEs

which I've implemented in the `GlyphSpanner`.

On another note, I think it might be possible to use the same marked content for ReversedChars and ActualText in case we have both, but I'm not sure whether that's advisable or whether it'll just confuse viewers ...

I added a snapshot test `text_direction_auto`. However, that test is a copy of `font_arabic_simple_text`, which is a visreg test. I couldn't figure out whether it's possible to have one test be both. Right now, both tests are in different files which makes code sharing a bit weird, but if you'd prefer for me to rearrange this somehow, I'm happy to.


crates/krilla/src/content.rs
```diff
@@ -644,23 +644,32 @@ impl ContentBuilder {
                 let mut cur_x = x;
                 let mut cur_y = ys;
 
+                // When the glyph ranges are monotonically decreasing, the glyphs are in reverse
+                // order and we should write `/ReversedChars` marked content.
+                let reversed = glyphs.len() > 1
+                    && glyphs
+                        .windows(2)
+                        .all(|w| w[0].text_range().start >= w[1].text_range().end);
+
```
LaurenzV, line 653: I'm wondering whether it makes more sense to check if any of them is >? If I'm understanding correctly, in case we have glyphs that all map to the same text, we get a reversed by default, right?

PR #266, opened 2025-10-02
Bump hayro-write

saecki:
This bumps `hayro-write` to a slightly newer commit, in which `hayro` also depends on `skrifa 0.33`.
At some point `skrifa`, `parley`, etc. should probably be upgraded to the newest versions. But the roadblock I currently hit was that `svgtypes` and `usvg` still depend on `kurbo 0.11`.


laurmaedje:
doesn't hayro already depend on kurbo 0.12? I think it's not that bad if have kurbo twice anyway.

PR #271, opened 2025-10-06
Write authors as array in PDF 2.0

reknih:
Currently, authors are always folded into a comma-separated single entry in XMP, even though it supports an array. This is a compatibility behavior. The background behind this is that the PDF Document Info dictionary cannot express multiple authors but XMP can. When a document has both present and they do not match, Acrobat will first render the DocInfo author and then the 2nd , 3rd, ... author from the XMP.

In PDF 2.0, the whole document info dictionary has been deprecated in favor of XMP. That means that for these documents, we can disable this compatibility behavior and write an array.

### DIFF
--- a/crates/krilla-tests/src/main.rs
+++ b/crates/krilla-tests/src/main.rs
@@ -1077,6 +1077,14 @@ pub fn settings_29() -> SerializeSettings {
     SerializeSettings::default()
 }
 
+pub fn settings_30() -> SerializeSettings {
+    SerializeSettings {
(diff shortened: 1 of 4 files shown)


LaurenzV:
Do you think you could add a simple test case for this?

LaurenzV:
I guess this also means we shouldn't write the information dictionary at all for 2.0 (in case we don't already), but that's for a future PR.

reknih:
`CreationDate` and `ModDate` are still allowed, so we are in the green.

PR #273, opened 2025-10-07
Complete artifact types

reknih:
This commit adds the PDF 1.7 artifact type `Background`, the PDF 1.7 pagination artifact subtype `Watermark`,  and the PDF 2.0 pagination artifact subtypes `PageNum`, `LineNum`, `Redaction`, `Bates`. Finally, the variant `PaginationOther` allows writing a pagination artifact without a subtype.

Because these new variants are not compatible with PDF 1.4, where the mechanism was introduced, a compatibility layer has been introduced.

### DIFF
--- a/crates/krilla/src/content.rs
+++ b/crates/krilla/src/content.rs
@@ -141,7 +141,10 @@ impl ContentBuilder {
             properties.pairs([(Name(b"MCID"), mcid)]);
         }
 
-        tag.write_properties(sc, properties);
+        // Page height extracted from transform and passed to function to allow
+        // its dependants to flip the y-axis, mirroring Krilla conventions.
+        let page_height = self.root_transform.ty();
+        tag.write_properties(sc, properties, page_height);
     }
 
     pub(crate) fn end_marked_content(&mut self) {
--- a/crates/krilla/src/interchange/tagging/mod.rs
+++ b/crates/krilla/src/interchange/tagging/mod.rs
@@ -131,7 +131,7 @@ use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashMap};
 use std::io::Write as _;
 
-use pdf_writer::types::{ArtifactSubtype, RoleMapOpts, StructRole, StructRole2};
+use pdf_writer::types::{RoleMapOpts, StructRole, StructRole2};
 use pdf_writer::writers::{PropertyList, StructElement};
 use pdf_writer::{Chunk, Finish, Name, Ref, Str, TextStr};
 use smallvec::SmallVec;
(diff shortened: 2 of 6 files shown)


LaurenzV:
Maybe we could again add a small test case that uses the a subtype?

reknih:
Both done. I was unsure whether the rectangle need to be transformed in any way.

saecki:
Yes it needs to be transformed, should be the same as here:
https://github.com/LaurenzV/krilla/blob/fe181c18783443f5801a83cf36c57296c17b97a2/crates/krilla/src/interchange/tagging/mod.rs#L826

saecki:
Oh, maybe it doesn't need any transform, since the page root transform is applied to the surface?

saecki:
I'm pretty sure it doesn't need to be transformed, but since it uses the same `Rect` type as the `bbox` attribute we should probably document that the `Rect` in the tag `bbox` is specified within the page transform, while the `bbox` for attributes is specified within the current `Surface` transform.

reknih:
> But since it uses the same `Rect` type as the `bbox` attribute we should probably document that the `Rect` in the tag `bbox` is specified within the page transform.

Done in https://github.com/LaurenzV/krilla/pull/273/commits/3fb3d2874944e947903defb99a10b4acaa9a2a03

refs/snapshots/tagging_artifact_subtypes.txt
```diff
@@ -0,0 +1,101 @@
...
+>>
+endobj
+
+4 0 obj
+<<
+  /Length 470
+>>
+stream
+/Artifact <<
+  /BBox [0 80 200 130]
```
LaurenzV: Is this really the correct bbox? If I understood correctly, we need to assume the default user space (i.e. where in PDF Coordinates it will actually end up being). Since the text is written at the top but the PDF coordinate system is y-down, shouldn't the y be something like 600 or 700 instead?

crates/krilla/src/interchange/tagging/mod.rs
```diff
@@ -147,26 +148,124 @@ pub use tag::*;
...
+/// An artifact that should not be part of the accessible structure.
+#[derive(Copy, Clone, Debug, PartialEq, Default)]
+pub struct Artifact {
+    /// The type of the artifact.
+    pub kind: ArtifactType,
+    /// The bounding box of the artifact. Required for background artifacts.
+    ///
+    /// Unlike the [attribute `BBox`](crate::tagging::Tag::bbox) on non-artifact
+    /// structure elements, this property list entry is specified with regards
+    /// to the page transform, not the [`Surface`](crate::surface::Surface)
```
LaurenzV: Are you sure about this? Does that mean you think it should only take the `y-flip` into account? I don't see why this would be the case, given the description in the spec.

crates/krilla/src/interchange/tagging/generated.rs
```diff
@@ -4384,6 +4428,10 @@ pub(crate) enum LayoutAttr {
     WritingMode(WritingMode),
     /// The bounding box of a tag that encloses its visible content.
     /// If the content spans multiple pages, this should be omitted.
+    ///
+    /// Unlike the property list entry `BBox` on [artifacts](crate::tagging::Artifact),
```
LaurenzV: Why is this bbox different? As far as I can tell, in the tagging section of the spec it has the same description where it says in default user space units", so I don't see why it's different.

Image

LaurenzV (commented):
Overall LGTM, but we will have to do some more research on the whole bbox thing and then figure out how to best document how the bbox should be determined.

Unfortunately I don't have Acrobat Pro, so if someone else has it might be useful to test how Acrobat interprets the bbox attributes, especially if there are CTM operations in the main page stream.

crates/krilla/src/interchange/tagging/generated.rs
saecki: I'm sorry, I don't know how I got there. I've read through the spec again and that indeed doesn't make any sense.

Regarding the implementation, we could get the page transform from the `root_transform` of the `Surface::root_builder` in `Surface::start_tagged` and pass it through `ContentBuilder::start_marked_content_with_properties`.

Another idea I had, though that would be a larger change, would be to have a `SurfaceKind` with a `Page` variant that contains the page dimensions.
In the future we could also add a mutable borrow of the page annotations, so annotations could be added through the `Surface` which would come in handy when constructing a `TagTree`.

crates/krilla/src/interchange/tagging/mod.rs
saecki: No, you're right. I guess both the comment here and for the artifact could briefly state that the `Rect` is specified within the page transform.

refs/snapshots/tagging_artifact_subtypes.txt
saecki: I couldn't find a way to verify this using Acrobat or other tools, but I agree.

crates/krilla/src/interchange/tagging/generated.rs
LaurenzV: No worries! The only problem is that I've tried my best to abstract away the y-flip transform needed for PDF so that the user can always just assume a y-down coordinate system, it would be unfortunate if the user somehow has to take it into account here. Therefore, I think the best thing to do is to simply document that the bbox needs to cover all shapes in user space, but still allow the user to assume that user space is a y-down coorindate system. Then, internally we apply the y-flip transform so that it's converted to y-up, but this is hidden away from the user. WDYT?

saecki: I agree, the Y-flip should definitely stay hidden away, that's what I meant to say :)
That would require having access the page height to compute the transform right?
The implementation ideas were just two options I could think of to achieve that.

LaurenzV: Yep, but I’m pretty sure this should already exist somewhere in the code? I think we have something similar for annotation rects.

LaurenzV: Ah, but the difference is that those are written directly into the content stream… so yeah if there is an easy way to wire the page height through, that’s what we should do. I don’t currently remember if the automatic page size might cause problems here…

saecki: Is there a way to specify an automatic page size? Seems to me like the page size must be specified when creating a page :thinking: 
The surface also already uses the page height from the page settings when creating the root content builder, so that would match the behavior

LaurenzV: Then I must remember wrong! I’m sure there was this functionality in the past but I might have removed it already. 🤔

reknih: Yeah I must have missed that they are both defined in user-space units. Under this assumption, here is the new `BBox` rectangle from the test ImageMagick'ed into there. Neither Acrobat nor PAC 2026 were able to display the BBox.

tagging_artifact_subtypes_tight

LaurenzV:
LGTM, thanks!

crates/krilla/src/content.rs
```diff
@@ -141,7 +141,10 @@ impl ContentBuilder {
             properties.pairs([(Name(b"MCID"), mcid)]);
         }
 
-        tag.write_properties(sc, properties);
+        // Page height extracted from transform and passed to function to allow
+        // its dependants to flip the y-axis, mirroring Krilla conventions.
+        let page_height = self.root_transform.ty();
```
saecki, line 146: Is this really correct? This function will be called on the last `sub_builder`, which doesn't necessarily have the page root transform right?
I think to be correct we should use the `root_transform` of the `root_builder` inside `Surface::start_tagged`.
Also instead of extracting the `page_height` we could just pass through the entire transform.

PR #274, opened 2025-10-07
Always create CMap entries in PDF/A-1

reknih:
So far, we interpreted the PDF/A-1 spec to say that there must be a ToUnicode map, but not necessarily that it must contain every character. When testing the Typst test suite against veraPDF with PDF/A-1a, 8 complex shaping tests and 52 math tests failed its [rule 6.3.8-1](https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFA-Part-1-rules#rule-638-1), which is defined on the glyph level: They interpret the sentence "The font dictionary shall include a ToUnicode entry whose value is a CMap stream object that maps character codes to Unicode values" to say all glyphs must be mapped.

What neither the standard nor their test suite has is the rule preventing to map to 0, U+FFFE, or U+FEFF. Hence, I have added a new ValidationError variant NoCodepointMapping that means that no codepoint was found whereas InvalidCodepointMapping now only means that we did not like the codepoint we did find. They are enabled for the right standards.

The PR also forbids Tofus in A-1a for the same reason: Because they produce glyphs that are not mapped in the ToUnicode map, they are not allowed. A-1b is unaffected.

With these changes apply, all of the PDFs in Typst test suite that build in PDF/A-1a also pass veraPDF.

### DIFF
--- a/crates/krilla-tests/src/validate.rs
+++ b/crates/krilla-tests/src/validate.rs
@@ -17,8 +17,9 @@ use krilla_macros::snapshot;
 use crate::embed::{embedded_file_impl, file_1};
 use crate::{
     blue_fill, cmyk_fill, dummy_text_with_spans, green_fill, load_jpg_image, load_png_image, loc,
-    metadata_1, rect_to_path, red_fill, settings_13, settings_15, settings_19, settings_23,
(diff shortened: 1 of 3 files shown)


LaurenzV:
Thanks!

PR #275, opened 2025-10-07
Write `/Tabs S` even when page contains only links

reknih:
So far, a page needed tagged content to receive the `/Tabs S` entry in its page dictionary if it had an annotation. This PR instead always writes it when tagging is enabled and an annotation is present.

This addresses a veraPDF validation issue for PDF/UA-1 Clause 7.18.3 (Test 1) for the test `link-empty-block`.


LaurenzV:
(You'll need to update the snapshot,  as I was using main)

reknih:
Got it, sorry for the hassle

PR #276, opened 2025-10-07
Test whether veraPDF CI works

LaurenzV:


LaurenzV:
@reknih Could you add that test case to your PR? Thanks!

PR #277, opened 2025-10-08
Raise Missing*AltText error when alt text is empty

reknih:
This aligns with the [veraPDF rule](https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFUA-Part-1-rules#rule-73-1) to check for the presence of alt text

### DIFF
--- a/crates/krilla/src/interchange/tagging/mod.rs
+++ b/crates/krilla/src/interchange/tagging/mod.rs
@@ -711,10 +711,11 @@ impl TagGroup {
             *note_id += 1;
         }
 
-        if self.tag.can_have_title() && tag.title().is_none() {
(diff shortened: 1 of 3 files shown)


crates/krilla/src/interchange/tagging/mod.rs
```diff
@@ -712,9 +712,13 @@ impl TagGroup {
         }
 
         if self.tag.can_have_title() && tag.title().is_none() {
-            sc.register_validation_error(ValidationError::MissingHeadingTitle);
+            match tag.title() {
+                Some(title) if !title.is_empty() => {}
+                _ => sc.register_validation_error(ValidationError::MissingHeadingTitle),
+            }
```
LaurenzV, line 718: I think it should work here as well no? 😄

PR #278, opened 2025-10-08
Check that features required by UA-1 are available in the current PDF version

reknih:
This PR checks that three features are available by the current PDF version in circumstances in which PDF/UA-1 requires them. If they are needed but unavailable, a validation error is raised.

An earlier implementation of this PR just forbid export to these lower versions entirely.

### DIFF
--- a/crates/krilla-tests/src/validate.rs
+++ b/crates/krilla-tests/src/validate.rs
@@ -1,6 +1,7 @@
 use krilla::action::LinkAction;
 use krilla::annotation::{Annotation, LinkAnnotation, Target};
 use krilla::color::{rgb, separation};
+use krilla::configure::validate::VersionedFeature;
 use krilla::configure::ValidationError;
 use krilla::embed::EmbedError;
 use krilla::error::KrillaError;
@@ -19,9 +20,9 @@ use krilla_macros::snapshot;
 use crate::embed::{embedded_file_impl, file_1};
 use crate::{
     blue_fill, cmyk_fill, dummy_text_with_spans, green_fill, load_jpg_image, load_png_image, loc,
-    metadata_1, rect_to_path, red_fill, settings_13, settings_15, settings_19, settings_20,
-    settings_23, settings_24, settings_32, settings_7, settings_8, settings_9,
-    stops_with_2_solid_1, validation_errors, youtube_link, NOTO_SANS,
+    metadata_1, metadata_2, rect_to_path, red_fill, settings_13, settings_15, settings_19,
+    settings_20, settings_23, settings_24, settings_32, settings_33, settings_34, settings_7,
+    settings_8, settings_9, stops_with_2_solid_1, validation_errors, youtube_link, NOTO_SANS,
 };
 use crate::{Document, SerializeSettings};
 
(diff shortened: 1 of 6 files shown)


LaurenzV:
After thinking a bit more about this a bit more, I think the best thing to do is to simply only allow PDF/UA-1 with 1.7, even if in theory lower versions should be supported, as previously suggested. Any objections against this? @reknih @laurmaedje If not I'll open a PR for this.

reknih:
I'd be curious to know where the change of mind comes from: You have rightly asserted above that some files from older versions comply with UA-1. IMO raising a validation error only if the user requests a feature that is not possible with the Validator / PDF version tuple is elegant. Are you concerned about this behavior being unintuitive to users, difficult to document, or to maintain?

crates/krilla-tests/src/validate.rs
```diff
@@ -930,3 +931,117 @@ fn validate_deduplicate_errors() {
...
+
+    document.set_metadata(metadata_2());
+
+    let outline = Outline::new();
+    document.set_outline(outline);
+
+    assert_eq!(
+        document.finish(),
+        Err(KrillaError::Validation(vec![
+            ValidationError::RequiresLaterPdfVersion(
```
LaurenzV: Maybe `RequriesNewerPdfVersion` would be a bit better?

crates/krilla-tests/src/validate.rs
```diff
@@ -930,3 +931,117 @@ fn validate_deduplicate_errors() {
...
+                PdfFeature::HeaderFooterArtifactSubtypes,
```
LaurenzV: Maybe `VersionedFeature`? `PdfFeature` seems a bit too generic for me.

crates/krilla/src/configure/validate.rs
```diff
@@ -342,6 +366,7 @@ impl Validator {
                 ValidationError::MissingTagging => *self == Validator::A1_A,
                 ValidationError::MissingDocumentDate => true,
                 ValidationError::EmbeddedPDF(_) => true,
+                ValidationError::RequiresLaterPdfVersion(_, _) => false,
```
LaurenzV: I think we should explicitly spell out the three variants in all of these branches, so in case there is a new variant which requires validation in other standards, we don't overlook it.

crates/krilla/src/interchange/tagging/mod.rs
```diff
@@ -605,6 +618,41 @@ impl Node {
             },
         }
     }
+
+    // For tables to be valid in PDF versions before 1.5, we need to ensure that
+    // each table header cell (`TH`) is inside a table header (`THead`).
+    // Otherwise, the unsupported `Scope` attribute would be required to express
+    // the correct semantics.
+    fn validate_table_structure_before_pdf15(&self, in_header: bool, sc: &mut SerializeContext) {
```
LaurenzV: Hmm... Can we remove this for now and add a TODO instead? I do agree it would be nice to have tag tree validation, but I want to do it properly so that we can validate other constraints as well (for example the oens outlined in WCAG). Would like to avoid adding a method just for this specific case now in this PR.

crates/krilla/src/interchange/tagging/mod.rs
```diff
@@ -719,6 +767,12 @@ impl TagGroup {
             sc.register_validation_error(ValidationError::MissingAltText(tag.location));
         }
 
+        if matches!(self.tag, TagKind::Table(_)) && pdf_version < PdfVersion::Pdf15 {
+            for child in &self.children {
+                child.validate_table_structure_before_pdf15(false, sc);
+            }
```
LaurenzV: Same as above.

crates/krilla/src/page.rs
```diff
@@ -424,11 +425,15 @@ impl InternalPage {
...
-            page.tab_order(TabOrder::StructureOrder);
+        if !self.annotations.is_empty() && sc.serialize_settings().enable_tagging {
+            if sc.serialize_settings().pdf_version() >= PdfVersion::Pdf15 {
+                page.tab_order(TabOrder::StructureOrder);
+            } else {
+                sc.register_validation_error(ValidationError::RequiresLaterPdfVersion(
+                    PdfFeature::StructureOrderTabbing,
+                    sc.location,
+                ));
+            }
```
LaurenzV, line 457: Am I understanding correctly that this will fail any PDF 1.4 export with tagging enabled as soon as there is at least one annotation. Since the above comment suggests it's only needed for PDF/UA, let's only raise this error for that mode? Adding a simple `require_structure_order_tabbing` method to `Validators` should be enough. Let's also add a PDF 1.4 test with tagging and an annotation just to ensure that export succeeds (not a snapshot test, just check that `finish`.is_ok) so that this regression is covered.

LaurenzV (commented):
Overall LGTM, with comments addressed (and rebased onto newest main).

crates/krilla/src/page.rs
reknih: Since only PDF/UA-1 prohibits `ValidationError::RequiresNewerPdfVersion(VersionedFeature::StructureOrderTabbing, _)`, the error is filtered by `register_validation_error` for all other export targets. I have still added a regression test.

LaurenzV: Oops, you are right!

LaurenzV:
Happy to merge it once rebased.

reknih:
Thanks!

PR #283, opened 2025-10-17
Add `location` and `set_location` methods to `Tag<T>`

saecki:
This was an oversight, the methods were missing because previously the `location` field could be directly accessed. Now that `Tag<T>` wraps `AnyTag` it can't.


LaurenzV (approved):
LGTM otherwise, feel free to merge yourself.

PR #289, opened 2025-11-07
Allow writing empty table headers array

saecki:
I'm in the process of generating the serialized tag tree from the finished PDF document using `hayro` (nice API by the way!). I noticed that an empty table cell `Headers` attribute isn't written. I thought I removed this check but apparently I didn't.
From my understanding the `Headers` attribute allows overwriting the automatic header cell assignment based on `TH` cells and the `Scope` attribute, so writing an empty array would remove any associated header from a cell.

On another note how do you feel about the code in `tagging/fmt.rs`, it wouldn't be needed for Typst per se, but it might still be nice to print a tag tree?


saecki:
> Can we add a test for this?

Done

> Don't you guys already have tests for tag trees? Or what's the motivation here, out of curiosity.

I'm updating the test suite in order to also test PDFs and SVGs, the way we're currently testing the tag tree we would have to run the PDF export twice, once for the actual PDF and once for the  tag tree. In theory we could have some feature flag to return both from the function, but that all feels a little unclean. Currently we have a `#[doc(hidden)]` public function in `typst_pdf` that returns the formatted tag tree. It also allows reading some properties that are not possible within `krilla` without storing additional data that isn't needed otherwise. For example we can read the `Lang`, `Alt`, `E`, and `ActualText` attributes of marked content sequences and the `Contents` attribute of link annotations.
This is the current state: https://github.com/typst/typst/blob/test-suite/tests/src/pdftags.rs

PR #301, opened 2025-11-14
Add support for link borders

akretz:
I want to get typst to display boxes around links which are shown on screen but not printed. The inspiration comes from how LaTeX's hyperref package does it. So I looked into krilla and found a way to do just that. This issue has come up a few times and the suggested workarounds to achieve that all wrap links inside typst boxes:
https://forum.typst.app/t/how-to-add-color-box-to-a-non-citation-reference/3735
typst/typst#2585

These workarounds have the drawbacks that the boxes also appear on printed documents and it's probably not good from an accessibility point of view either. With this PR, it is possible to add borders around links according to the PDF standard just like hyperref does.

I have implemented the logic such that the print flag (which controls if the boxes are printed or not) is not set in case we do not obey to one of the PDF/A standards (because they require print output to be the same as screen output). I have set the `LinkBorder` from the typst PDF generator and get a hyperref-like behavior on my documents with this now.

I have put the border struct to the `LinkAnnotation` struct, but I'm not sure if it should rather belong to the `Annotation` struct, as any kind of annotation can have borders. I have done it this way to keep the PDF output unchanged.

### DIFF
--- a/crates/krilla-tests/src/annotation.rs
+++ b/crates/krilla-tests/src/annotation.rs
@@ -28,7 +28,7 @@ fn annotation_with_border(page: &mut Page) {
             Rect::from_xywh(50.0, 50.0, 100.0, 100.0).unwrap(),
             Target::Action(LinkAction::new("https://www.youtube.com".to_string()).into()),
         )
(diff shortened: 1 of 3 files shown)


LaurenzV:
Hi, thanks for the PR, I will take a look next week!

Issue #302, opened 2025-11-15
Proper handling of ligatures when creating tagged PDF


mwcampbell:
I thought I'd try modifying the parley example to output a tagged PDF. I did the simplest thing that I thought might be correct, putting the whole thing in a `P` tag in the tag tree, while using a `Span` content tag for each run. I used `SpanTag::empty()`. Everutying came out right, except that the word "fits" came out as "fts" when reading the tagged document using either Acrobat or `pdfinfo -struct-text`. No doubt this is because of the "fi" ligature.

So I think we need clarification on whether it's the application's responsibility to provide an `actual_text` for each `Span` content tag, or if Krilla is supposed to be smart enough to do this automatically. Krilla is clearly already creating some `Span` content tags with `ActualText` for the complex scripts, just not for this Latin ligature. So is that by design, or an oversight?

LaurenzV:
Hey, thanks for trying it. I should have noted somewhere that the parley example is a bit broken. I tried printing out the glyphs that are generated by parley:

```
KrillaGlyph { glyph_id: GlyphId(87), text_range: 63..64, x_advance: 0.36100003, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, t
KrillaGlyph { glyph_id: GlyphId(75), text_range: 64..65, x_advance: 0.61800003, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, h
KrillaGlyph { glyph_id: GlyphId(68), text_range: 65..66, x_advance: 0.56100005, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, a
KrillaGlyph { glyph_id: GlyphId(87), text_range: 66..67, x_advance: 0.36100003, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, t
KrillaGlyph { glyph_id: GlyphId(3), text_range: 67..68, x_advance: 0.26000002, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None },  
KrillaGlyph { glyph_id: GlyphId(76), text_range: 68..69, x_advance: 0.25800002, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, i
KrillaGlyph { glyph_id: GlyphId(87), text_range: 69..70, x_advance: 0.36100003, x_offset: 0.0, y_offset: 0.0, y_advance: 0.0, location: None }, t
[...]

PR #303, opened 2025-11-15
Update Parley example

mwcampbell:
Parley has been through several releases since the Git revision that the example is currently using. So this update needs to account for a few breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,7 +557,8 @@ dependencies = [
 [[package]]
 name = "fontique"
 version = "0.6.0"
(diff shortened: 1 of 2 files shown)


mwcampbell:
I figured I should do this before I try adding a `text_range` field to Parley's `Glyph` to address #302.

LaurenzV:
Does this correspond to 0.6? If so I think we can just straight up use the crates.io version, no?

LaurenzV:
Thanks a lot!

PR #304, opened 2025-11-16
Sort name entries by key

LaurenzV:
See https://github.com/typst/typst/issues/7248.

### DIFF
--- a/crates/krilla/src/chunk_container.rs
+++ b/crates/krilla/src/chunk_container.rs
@@ -259,7 +259,7 @@ impl ChunkContainer {
                     // self-consistent; keys shall be compared for equality on
                     // a simple byte-by-byte basis."
                     let mut sorted = named_destinations.into_iter().collect::<Vec<_>>();
-                    sorted.sort_by(|a, b| a.0.name.as_bytes().cmp(&b.0.name.as_bytes()));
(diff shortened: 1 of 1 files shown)


LaurenzV:
Tested this with my master's thesis and it seems to fix the link issues in Preview.

crates/krilla/src/chunk_container.rs
```diff
@@ -250,9 +250,16 @@ impl ChunkContainer {
...
+                    // contain the tree’s keys and their associated values,
+                    // arranged in key-value pairs and shall be sorted lexically
+                    // in ascending order by key. Shorter keys shall appear
+                    // before longer ones beginning with the same byte sequence.
+                    // Any encoding of the keys may be used as long as it is
+                    // self-consistent; keys shall be compared for equality on
+                    // a simple byte-by-byte basis."
                     let mut sorted = named_destinations.into_iter().collect::<Vec<_>>();
-                    sorted.sort_by(|a, b| a.1.cmp(&b.1));
+                    sorted.sort_by(|a, b| a.0.name.as_bytes().cmp(b.0.name.as_bytes()));
```
laurmaedje: I think Typst won't trigger this because it deduplicates itself, but technically could it happen that we have two named destinations with the same name here and then the stable sort will retain the non-deterministic iteration order of the hashmap, leading to reproducibility problems?

PR #305, opened 2025-11-16
Fix ligature text ranges in the Parley example

### LINKED ISSUE #302: Proper handling of ligatures when creating tagged PDF

mwcampbell:
Fixes #302 without modifying Parley.

### DIFF
--- a/crates/krilla/examples/parley.rs
+++ b/crates/krilla/examples/parley.rs
@@ -87,9 +87,17 @@ fn main() {
             // current glyphs into the PDF, and build the next sequence of consecutive
             // glyphs.
             let mut cur_style = None;
-            let mut glyphs = vec![];
+            let mut glyphs = Vec::<KrillaGlyph>::new();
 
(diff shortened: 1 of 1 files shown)


LaurenzV:
Seems to work, thanks!

PR #312, opened 2025-11-27
Use `IndexMap` to sort fonts by usage order

saecki:
Use `IndexMap` to sort fonts by usage order. If the order isn't deterministic, the PDF won't be deterministic anyway.
The hashing approach previously used produces different orders on my local machine and a github actions runner.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,6 +783,7 @@ dependencies = [
  "image 0.25.2",
  "image-webp",
  "imagesize 0.14.0",
+ "indexmap",
  "once_cell",
  "oxipng",
  "parley",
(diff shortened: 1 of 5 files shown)


saecki:
So there are a couple of things here:
- The `SipHashable` implementation of `Prehashed` also hashes the `type_id`, which doesn't have any stability guarantees between compilation cycles.
  - This breaks the sort order of fonts.
- The default Hash implementation of slices writes a length prefix as a usize to any hasher.
  - This causes the differences on 32-bit platforms, I've implemented an easy fix for that here: https://github.com/LaurenzV/krilla/commit/ced12b64acf3536f2e11e08762ca7048c5b79ec8
  (But that's not relevant to this PR)

laurmaedje:
Then, this is probably good to go?

LaurenzV:
Is there any way we can feasibly test this in CI? We actually already have a test with 2 fonts that is compiled repeatedly to check it says stable, but that doesn't seem to be affected by the bug.

PR #315, opened 2025-12-02
Remove unused `font_cache`

saecki:
Seems like this was a left over from refactoring.


LaurenzV:
Ah yeah, they are cached within krilla-svg now...

PR #323, opened 2025-12-27
Add Separation color space

reknih:
This PR adds support for the special `Separation` color space. It exposes the colorant name in the `Separation` dictionary, including typed writing for the `All` and `None` colorants. For fallback, the color space accepts a color that is then linearly interpolated given the tint, so both the `alternateSpace` and `tintTransform` array entries are managed by Krilla.

On standards compliance: PDF/A-2 and later require the same colorants to always come with the same `alternateSpace` and `tintTransform` entries. I am tracking this in the new `ValidationStore` struct. PDF/UA has no special requirements for separation color spaces.

I had to remove the `Copy` impl from Color because Separations can contain strings. Let me know if e.g. `separation::Color` should contain a reference instead or `SeparationColorant` a Cow instead of a plain string.

I have added snapshot tests and manually inspected the result in a PDF reader.

### DIFF
--- a/crates/krilla/src/content.rs
+++ b/crates/krilla/src/content.rs
@@ -970,7 +970,7 @@ impl ContentBuilder {
         opacity: NormalizedF32,
         sc: &mut SerializeContext,
         mut set_pattern_fn: impl FnMut(&mut Content, String),
-        mut set_solid_fn: impl FnMut(&mut Content, ContentColorSpace, Color),
+        mut set_solid_fn: impl FnMut(&mut Content, ContentColorSpace, &Color),
     ) {
         let pattern_transform = |transform: Transform| -> Transform {
             transform.post_concat(self.cur_transform_with_root_transform())
@@ -1023,7 +1023,7 @@ impl ContentBuilder {
             InnerPaint::Color(c) => {
                 let cs = c.color_space(sc);
                 let color_space_resource = Self::cs_to_content_cs(self, sc, cs);
-                set_solid_fn(&mut self.content, color_space_resource, c.clone());
+                set_solid_fn(&mut self.content, color_space_resource, c);
             }
             InnerPaint::LinearGradient(lg) => {
                 let (gradient_props, transform) = lg.clone().gradient_properties(bounds);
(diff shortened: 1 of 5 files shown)


crates/krilla/src/content.rs
```diff
@@ -1087,21 +1087,24 @@ impl ContentBuilder {
...
                         content.set_fill_cmyk(comps[0], comps[1], comps[2], comps[3]);
                     }
+                    Color::Special(_) => {
+                        panic!("Device color space cannot be used with special colors")
+                    }
                 },
                 ContentColorSpace::Named(n) => {
                     content.set_fill_color_space(n.to_pdf_name());
-                    content.set_fill_color(color.to_pdf_color());
+                    content.set_fill_color(color.clone().to_pdf_color());
```
saecki: I think this `clone()` can be removed

crates/krilla/src/graphics/color.rs
```diff
@@ -246,32 +347,203 @@ pub mod rgb {
...
+    }
+}
+
+/// Separation (spot) colors.
+pub mod separation {
+    use crate::color::RegularColor;
+
+    /// A spot color.
+    #[derive(Debug, Hash, Eq, PartialEq, Clone)]
+    pub struct Color(pub(crate) u8, pub(crate) SeparationSpace);
```
saecki: I think a regular struct would be more readable than a tuple struct

crates/krilla/src/graphics/color.rs
```diff
@@ -65,33 +92,83 @@ pub enum Color {
...
+                    DeviceColorSpace::Cmyk.into()
                 }
                 Some(cs) => cs,
             },
         }
     }
+
+    /// Return the current color as RGB for use with colored glyphs (SVG and
+    /// COLR).
+    pub(crate) fn as_rgb(self) -> Option<rgb::Color> {
```
saecki: The name `to_rgb` would be more rust idiomatic, but I see this is in line with the `as_rgb` function for `Paint`. Mabye we could change that as well, what do you think @LaurenzV?

crates/krilla/src/graphics/shading_function.rs
```diff
@@ -89,12 +89,12 @@ impl GradientProperties {
         match self {
             GradientProperties::RadialAxialGradient(rag) => {
                 if rag.stops.len() == 1 {
-                    return Some((rag.stops[0].color, rag.stops[0].opacity));
+                    return Some((rag.stops[0].color.clone(), rag.stops[0].opacity));
```
saecki: With `set_solid_fn` taking a reference, this can also just return a reference to the `Color`

saecki (commented):
The general changes look good! I've just got a few nitpicks :)

crates/krilla/src/graphics/color.rs
reknih: Thanks!

PR #324, opened 2026-01-03
Create Python bindings for krilla

jbarlow83:
Hi! I'm the author [OCRmyPDF](https://github.com/ocrmypdf/OCRmyPDF/) and I've been looking for a better solution for rendering text obtained from OCR engines back into PDF, and converged on Krilla. It seems to have real solid font and handling and PDF generation, and I'm exciting about being able to potentially generate tagged PDFs eventually.

With OCR, the text positioning generated by the OCR engine never matches what a typesetting engine wants to do - you have to manually stretch and reposition to achieve alignment to the iamge. I used Krilla to tackle a difficult, nonstandard problem and it did extremely well. (I am currently using a "Glyphless" font that has only a space character, combined with PDF Unicode tables that map every code point to that glyph, and a low level PDF content stream generator. It works, but it's a hack and many PDF viewers/text extractors struggle with word segmentation of the output or anything to do with RTL.)

Since OCRmyPDF is Python, I created Python bindings for krilla. I did not do krilla-svg but covered pretty much everything else. I thought it would be easiest to propose this for merging upstream since it would allow others to benefit and bring krilla to a wider user base. 

Most of the API is covered, with adjustments for Pythonic ergonomics in a few places:
- builder pattern -> kwargs
- support for Python str
- uharfbuzz because Python doesn't have parley
- Python context managers

Some fairly straightforward unsafe code is required to deal with Python's lifetime semantics, such as the ability to grab a reference to a child object and then delete the parent. So we hold any issued pointers in `DocumentState` object and don't let any of them go away until reference count drops to 0. The rationale is carefully documented. 

The CI is modified with code to upload to PyPI. You would need to register an account there, approve GitHub as a publisher to PyPI for that to work.

The diff of the krilla create is very small - a `rust clippy` issue (paint.rs) that may need to be fixed anyway, and exposing two `pub(crate)` functions. There are cargo tests for the rust side, pytest tests for Python, and brief Python docs with autoapi.

I did not attempt attempt to set up the visualization tests - the Python binding layer isn't altering any existing rendering functionality; it's just Krilla do what it already does.

### DIFF
--- a/crates/krilla-py/examples/pil_image.py
+++ b/crates/krilla-py/examples/pil_image.py
@@ -36,7 +36,10 @@ def main():
     # Create a PDF and embed the image
     doc = Document()
 
-    with doc.start_page_with(PageSettings.from_wh(450, 250)) as page, page.surface() as surface:
+    with (
+        doc.start_page_with(PageSettings.from_wh(450, 250)) as page,
+        page.surface() as surface,
+    ):
         # Draw the image at its full size
         surface.draw_image(krilla_img, Size.from_wh(400, 200))
 


LaurenzV:
Hi, thanks a lot for the PR, it's great to see the library potentially getting more adoption! :)

However, unfortunately I think I would rather not have those bindings in the main krilla repo. Don't get me wrong, I definitely think it would be great if krilla gets bindings to other languages, but if they live here, it becomes my responsibility to maintain them, and since I'm not at all familiar in that area, I would prefer not having to do that. What do you think if you were to just create your own repo with them (maybe by including krilla as a git submodule) so you can maintain them at your own pace? I think there is a lot of precedent for that with other crates, for example the [JavaScript bindings for resvg](https://github.com/thx/resvg-js) or [Python bindings for Typst](https://github.com/messense/typst-py), so that seems like the most desirable approach to me.

PR #331, opened 2026-02-02
Support embedding a number of PNGs directly 

andylizi:
Add initial support for directly embedding PNG files into PDF without re-encoding.

This is motivated by typst/typst#7668. Dedicated png optimizers like `oxipng` can often produce significantly smaller files, and it'd be nice if we can benefit from this. Skipping the re-encoding step is also more performant.

Directly-embedded PNGs are smaller due to the [filtering](https://en.wikipedia.org/wiki/PNG#Filtering) step, which makes the raw pixels more compressible. In contrast, krilla currently applies a Flate stream filter to the image data without any pre-processing, leading to suboptimal compression.

```diff
   /Type /XObject
   /Subtype /Image
   /Filter /FlateDecode
+  /DecodeParms <<
+    /Predictor 15        <-- Tell PDF viewers to decode PNG prediction after decompression
+    /Colors 3
+    /BitsPerComponent 8
+    /Columns 200
+  >>
   /Width 200
   /Height 200
```

### PNG Support

- **Implemented in this PR** (must meet all conditions):
  - Not interlaced
  - Not indexed color
  - No alpha channel
  - 8 or 16 bit depth
- **Can be implemented with follow-up work**:
  - 1, 2, and 4 bit depth
  - Indexed color
- **Cannot be implemented**:
  - Interlaced PNGs
- **Unsure**:
  - PNGs with alpha channel
    - Four components per sample *is* supported according to the spec, so I don't think PDF readers would outright reject `FlateDecode` filters with an alpha channel. But a SMask would be required for transparency anyway: would it really be better to embed the alpha data twice just to avoid re-encoding?

### Drawbacks

- Will inevitably produce larger files when the input PNG is *not* well-compressed. In my opinion, this could be an acceptable trade-off, because it gives people the option to optimize if they care about the file size. But without this capability, there is nothing they can do regardless.
  - Maybe make it configurable?
  - Maybe we can unconditionally re-encode first, compare the sizes, and take whichever method that is smaller. Sacrifice performance, but no worse than it is now.
- The `image-png` crate does not support extracting raw IDAT chunks, so I had to roll my own parser, adding complexity. Thankfully the format is very straightforward.
- If the raw data in PNG is corrupted, but the CRC32 checksum is still valid, we won't be able to discover the error. Very unlikely to happen accidentally.

### Correctness

- Implemented according to the PDF spec, specifically the table "Optional parameters for LZWDecode and FlateDecode filters"
- I've done manual, rudimentary sanity checks using PDFium, Adobe Acrobat, and pdf.js, but nothing comprehensive yet.
- I've compared our PDF output against [img2pdf](https://github.com/josch/img2pdf) to make sure they're similar. However, we cannot look at their LGPL-licensed implementation.

### Typst

Typst does not currently use the native PNG code path in krilla, so it won't be affected by this change. We can add support with a few lines of code.


TechPizzaDev:
I got here from the Typst issue, and have a case where my indexed PNGs grow by 8x when exported to PDF. Commenting as a bump but also question; Am I reading it right that adding support for `/Indexed` should be relatively straight-forward? 

I can't tell if the `png` crate would be useful in extracting the palette, but you could borrow [their palette parsing](https://github.com/image-rs/image-png/blob/95ac73859a2b5de7ec6dd72f2830683e52c1abdf/src/decoder/transform/palette.rs#L39-L91) into `png_raw`.

LaurenzV:
I presume indexed images could be implemented on top of this, yes. Unfortunately I still haven't found the time to properly review this PR. :/ I do still hope to get to it eventually....

reknih:
I would maybe add one additional restriction since there are a few non-critical chunks that can affect appearance:
`tRNS` is a secondary way to declare a color transparent even if color type is greyscale or truecolor. Since we currently forbid transparent images, the `bKGD` chunk can be ignored but may need to be considered for transparent images.

andylizi:
Hi all, before resolving the merge conflicts, I'd like some input on the general direction: should we always skip the re-encoding (current implementation), or should always re-encode first, compare the sizes, and take whichever method that is smaller?

I'm leaning toward the latter approach since there'd be fewer file size regressions.

LaurenzV:
Hi, please don’t worry about rebasing this PR, I will take care of this once I get to it! It is fully my fault that this PR has been hanging around for so long, so I will take care of taking this over the finish line once I find the time + motivation. 🙏

LaurenzV:
Apologies again for the delay, and thanks for pushing this forward. I have made some changes and added some tests, but it was a good basis to work on!

Some notes:
- I'm planning on adding support for indexed PNGs and PNGs with `tRNS` transparency in a follow-up. Not sure if this weekend though.
- Once I'm done with that, I'll also run a more exhaustive test with real-life PNGs to make sure there are no regressions.
- A disadvantage of this approach now is that we can't actually validate whether the PNG stream is correct, so if it's invalid you might end up embedding invalid data. We could decompress it to check this, but it would be unfortunate. However, I believe Typst always eagerly converts into a `DynamicImage`, so it might not be a problem? FYI @laurmaedje

A disadvantage is that if you want to use this in Typst right now, you'd probably have to always call this API. However, if the fast PNG path is not supported, then it will end up decompressing the image, which is unnecesasry because Typst already has an uncompressed version. Perhaps we need an additional API here so Typst can check whether the fast path is supported. Or Typst stops eagerly decompressing images.

PR #343, opened 2026-04-15
Allow specifying open/closed state on `OutlineNode`

mitsuru:
## Summary

Add `OutlineNode::new_with_state` and `OutlineNode::set_open` so callers can control whether a node's children appear expanded when the PDF is first opened. The existing `new` constructor keeps producing a *closed* node, so existing users see byte-identical output (the `outline_simple` snapshot is unchanged).

Internally, the `/Count` value is now computed by walking the subtree and summing only the items visible under each node's current open/closed state, as required by PDF 1.7 §12.3.3. For all-closed trees this matches the previous `children.len()`-based count, so existing behavior is preserved; for trees with mixed or open items it produces the correct visible-descendant count.

## Motivation

Downstream consumers that implement the CSS GCPM `bookmark-state: open | closed` property (e.g. Prince, WeasyPrint, and the HTML→PDF engine I'm working on) need a way to mark individual outline entries as initially expanded. Today krilla always emits a negative `/Count` for items with children, so every PDF viewer displays the outline fully collapsed.

## API

```rust
// unchanged — still produces a closed node (default)
OutlineNode::new(text, destination);

// new — explicitly specify the initial state
OutlineNode::new_with_state(text, destination, /* open = */ true);

// new — builder-style mutator
node.set_open(true);
```

No changes to `Outline` (root) — the PDF spec requires the `/Outlines` dictionary's `/Count` to be positive, and that is already the case.

## Test plan

- `cargo test -p krilla` — passes
- `cargo test -p krilla-tests outline` — existing `outline_simple` and `outline_with_empty_title` snapshots unchanged
- New snapshots:
  - `outline_open_state` — a top-level `Heading 1` with `set_open(true)` emits `/Count +1` (positive), and the root `/Outlines` `/Count` grows from 2 to 3 since the open grand-child is now visible
  - `outline_mixed_state` — a three-level tree with mixed open/closed parents; verifies that recursive visibility is counted correctly (open parent: `/Count +2`, closed parent: `/Count -1`, root: `/Count 4`)
- Manually verified the emitted PDFs in a PDF viewer: `outline_open_state.pdf` shows `Heading 1` initially expanded, `outline_mixed_state.pdf` shows the open/closed mix as expected
- `cargo fmt --check` — clean
- `cargo clippy --all-targets -p krilla -p krilla-tests` — no new warnings

### DIFF
--- a/crates/krilla/src/interchange/outline.rs
+++ b/crates/krilla/src/interchange/outline.rs
@@ -141,7 +141,7 @@ impl OutlineNode {
     /// `text` is the string that should be displayed in the outline tree, and
     /// `destination` is the destination that should be jumped to when clicking on
     /// the outline entry. The node is created in the *closed* state; use
-    /// [`OutlineNode::new_with_state`] or [`OutlineNode::set_open`] to change this.
+    /// [`OutlineNode::with_open`] or [`OutlineNode::set_open`] to change this.
     pub fn new(text: String, destination: XyzDestination) -> Self {
         Self {
             children: vec![],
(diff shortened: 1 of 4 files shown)


crates/krilla/src/interchange/outline.rs
```diff
@@ -126,22 +131,48 @@ pub struct OutlineNode {
...
+            text,
+            destination,
+            open,
         }
     }
 
+    /// Set whether this node is initially open (children shown expanded).
+    ///
+    /// Returns `&mut Self` to allow chaining with other builder-style calls.
+    pub fn set_open(&mut self, open: bool) -> &mut Self {
```
saecki, line 171: Krilla usually uses the `with_` prefix and methods that take `self` by value for its builders.
I guess the `set_` method without any return value could be provided as well...

mitsuru: @saecki 
Thanks, good point — I hadn't checked the surrounding conventions carefully. Updated in 11e0d94:

- Replaced new_with_state with with_open(mut self, open: bool) -> Self.
- Changed set_open to return () (matching the set_location / set_fill style).

The new doc comment and the outline_mixed_state test were updated to use the new builder form.

LaurenzV:
Applied some small tweaks but LGTM otherwise, thanks!

PR #345, opened 2026-04-17
Add `Document::finish_to_writer` for streaming PDF output

gpradofe:
## Motivation

`Document::finish()` returns a `Vec<u8>` containing the full PDF bytes. For large documents (hundreds of pages, heavy tables, embedded images, etc.) this forces the caller to hold the entire serialized PDF in memory before it can be piped to a file, a socket, or any other sink.

We've hit this in a Typst fork that generates big tabular PDFs: we need the final byte stream to go straight to a file without materializing another copy of the entire document on the heap right before `write_all`.

## What this PR adds

1. **`Document::finish_to_writer<W: std::io::Write>(self, writer: W) -> KrillaResult<()>`**

   Same semantics as `finish()`, but writes into any `io::Write` sink instead of returning a `Vec<u8>`. The byte stream is byte-identical to what `finish()` would return.

   Call-site before:
   ```rust
   let bytes = doc.finish()?;
   std::fs::write("out.pdf", bytes)?;
   ```

   Call-site after:
   ```rust
   let file = std::fs::File::create("out.pdf")?;
   doc.finish_to_writer(file)?;
   ```

2. **`KrillaError::Io(String)`** — a new error variant that surfaces any `io::Error` raised by the writer.

3. **Small refactor in `PdfVersion`**: extract a private `version_tuple()` helper from `set_version()`, so a future internal streaming-serializer path can look up the `(major, minor)` pair without needing a `Pdf` instance to mutate.

The scope is deliberately narrow. Internally `finish_to_writer` currently forwards the result of `finish()` into `write_all`; the public API is the thing that's useful today, and a follow-up can make the internal writes truly streaming (so the full buffer is never materialized) without changing this entry point.

## Tests

Added two unit tests in `crates/krilla/src/document.rs`:

- `finish_to_writer_matches_finish` — builds two identical documents, finishes one via `finish()` and the other via `finish_to_writer` into a `Vec<u8>`, asserts the byte streams are equal.
- `finish_to_writer_propagates_io_error` — a `Write` impl that always returns `ErrorKind::BrokenPipe` surfaces as `KrillaError::Io`.

### Test plan
- [x] `cargo test -p krilla --lib document` passes.
- [x] `cargo clippy -p krilla --lib --all-features` clean under `-Dwarnings`.
- [x] `cargo fmt --check` clean.

Happy to split the refactor of `version_tuple` out if you'd prefer — it's only used by the streaming entry point right now, but the entry point doesn't strictly need it yet, so I can drop it if it feels like noise.

### DIFF
--- a/crates/krilla/src/configure/version.rs
+++ b/crates/krilla/src/configure/version.rs
@@ -82,13 +82,21 @@ impl PdfVersion {
     }
 
     pub(crate) fn set_version(&self, pdf: &mut Pdf) {
+        let (major, minor) = self.version_tuple();
+        pdf.set_version(major, minor);
+    }
+
+    /// The `(major, minor)` numbers for this PDF version. Split out so it
+    /// can be used by streaming-serialize paths that don't have a `Pdf`
+    /// instance handy.
+    pub(crate) fn version_tuple(&self) -> (u8, u8) {
         match self {
-            PdfVersion::Pdf14 => pdf.set_version(1, 4),
-            PdfVersion::Pdf15 => pdf.set_version(1, 5),
-            PdfVersion::Pdf16 => pdf.set_version(1, 6),
-            PdfVersion::Pdf17 => pdf.set_version(1, 7),
-            PdfVersion::Pdf20 => pdf.set_version(2, 0),
-        };
+            PdfVersion::Pdf14 => (1, 4),
+            PdfVersion::Pdf15 => (1, 5),
+            PdfVersion::Pdf16 => (1, 6),
+            PdfVersion::Pdf17 => (1, 7),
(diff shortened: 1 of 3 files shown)


LaurenzV:
Sorry, but unless I'm missing something, that PR doesn't make any sense. You are still calling `self.serializer_context.finish()?.finish()`, which will still serialize this into a `Vec<u8>` beforehand. So this isn't going to help at all.

And even if that worked, I don't think there is much point in having such an API. Internally, krilla still needs to keep all objects fully allocated before writing them out, so the gains here would really be minimal if at all. You should be able to create PDFs with thousands of pages without running into major issues (unless of course everything is big images, but then such a change wouldn't really help much).

Therefore, I will close this. Sorry!

PR #352, opened 2026-04-18
Ability to use multiple validators

### LINKED ISSUE #101: Support multiple validators at once
Since a PDF file can conform to PDF/A and PDF/UA at the same time, multiple validators could be active at the same time. From an initial look at the validation implementation, that seems fairly doable. Most checks on a list of validators would just become a check whether any of the validators requires the feature. This would be necessary for Typst (at least in the long run) because we want to allow users to specify multiple (compatible) standards at the same time.

reknih:
This PR adds the ability to combine validators, in particular, it will enable the sought-after combination PDF/A-2a and PDF/UA-1.

Closes #101


reknih:
For the veraPDF test failure, see https://github.com/veraPDF/veraPDF-library/issues/1588

LaurenzV:
Thanks! As it happens, I was thinking about taking this task on myself. I want to spend some more time thinking about whether the validators can be represented a bit more elegantly, I might either push to this PR or create a new one with the tests you added!

saecki:
Not sure if you meant using a bitset for the validators, but I think that would be nice :)

reknih:
I considered eagerly collecting all validators into a shared policy object to a) only iterate them once and b) encapsulate how their policies fold with regards to any/all and negations.

However, since in practice the validator vector will have at most 2 items, the iterations are not too bad. Having the policy folding in one place would be nice but I eschewed that in pursuit of a minimal change (or laziness).

I don't know how useful a bit field would be since (at least with how policies are expressed now) ORing them does not yield a meaningful result.

LaurenzV:
I updated veraPDF in CI, so hopefully it should work now?

saecki:
> I don't know how useful a bit field would be since (at least with how policies are expressed now) ORing them does not yield a meaningful result.

The bitset would only be used internally to have a cheaply copyable type that can hold all standards. Though the proposed design in #355 from @LaurenzV would probably be nicer for this too.
The reason I'd want to have a cheaply copyable type would be, so that `ValidationError`s could include the standard/validator that caused the failure. With your proposed `ValidationPolicy` struct, each policy would then hold this struct and the validation error could just copy it. With the current design, it would be constructed ad-hoc instead of doing the `any`/`all` iteration over validators, basically doing a `filter_map().collect()`.

reknih:
I will attempt to build a `ValidationPolicy` approach over the course of the next ten days.

reknih:
I have updated the PR with a new design merging the three approaches, with a heavy dose of #355. See the commit messages for a more in-depth design rationale.

crates/krilla/src/configure/validate.rs
```diff
@@ -142,15 +142,294 @@ pub enum ValidationError {
...
+    }
+}
+
+/// A builder for constructing a [`Validators`] collection.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default, Hash)]
+pub struct ValidatorsBuilder(Validators);
+
+impl ValidatorsBuilder {
+    /// Set a validator (overwrites if same standard family already set).
+    pub fn set_validator(self, validator: Validator) -> Self {
```
saecki, line 316: I think more idiomatic names for these would be `with_validator` and `try_with_validator`

crates/krilla/src/configure/validate.rs
```diff
@@ -309,496 +517,647 @@ pub enum Validator {
...
-            Validator::A2_A => true,
-            Validator::A2_B | Validator::A2_U => false,
-            Validator::A3_A => true,
-            Validator::A3_B | Validator::A3_U => false,
-            Validator::A4 | Validator::A4F | Validator::A4E => false,
-            Validator::UA1 => true,
+/// A validator for exporting PDF documents to a specific subset of PDF.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[allow(non_camel_case_types)]
+pub enum Ua {
```
saecki: I know this is just taken from my example, but I think renaming this to `Accessibility` would be more consistent with the methods names.

crates/krilla/src/configure/mod.rs
```diff
@@ -3,70 +3,247 @@
...
+
+    /// Set the PDF/UA accessibility validator, returning Err if one is already set.
+    pub fn set_accessibility_validator_once(mut self, ua: Ua) -> Result<Self, Ua> {
+        self.validators = self.validators.set_accessibility_validator_once(ua)?;
+        Ok(self)
+    }
+
+    /// Build the [`Configuration`], returning an error if the validators and version are incompatible.
+    pub fn finish(self) -> Result<Configuration, ConfigurationBuilderError> {
+        match self.validators.finish() {
```
saecki: Replacing the outer match statement with this makes this a little easier to read:

```rs
let validators = self
    .validators
    .finish()
    .map_err(ConfigurationBuilderError::NoOverlappingValidatorsRange)?;
```

crates/krilla/src/configure/mod.rs
```diff
@@ -3,70 +3,247 @@
...
+            Err(validators) => Err(ConfigurationBuilderError::NoOverlappingValidatorsRange(
+                validators,
+            )),
+        }
+    }
+}
+
+/// An error that occurred while building a [`Configuration`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConfigurationBuilderError {
```
saecki: Maybe just `ConfigurationError`?

saecki (commented):
Nice, I think having the offending validators will save a lot of people a whole lot of frustration.
And sorry for the all the back and forth.
Just some final nitpicks from my side

PR #355, opened 2026-04-19
Add support for multiple validators

LaurenzV:
Still need to review the tests and other changes that were made, but thoughts on this API? @reknih

### DIFF
--- a/crates/krilla/src/configure/validate.rs
+++ b/crates/krilla/src/configure/validate.rs
@@ -142,15 +142,10 @@ pub enum ValidationError {
     EmbeddedPDF(Option<Location>),
 }
 
-/// A validator for exporting PDF documents to a specific subset of PDF.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
+/// The archival standard to export the PDF to.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[allow(non_camel_case_types)]
-pub enum Validator {
-    /// A dummy validator, that does not perform any actual validation.
-    ///
-    /// **Requirements**: -
-    #[default]
-    None,
+pub enum Archival {
     /// The validator for the PDF/A-1a standard.
     ///
     /// **Requirements**:
@@ -221,6 +216,27 @@ pub enum Validator {
     /// **Requirements**:
     /// - All requirements of PDF/A-2b
     A3_U,
+    /// The validator for the PDF/A-4 standard.
+    ///
+    /// **Requirements**:
+    /// - While not required, it's recommended to enable tagging.
+    A4,
+    /// The validator for the PDF/A-4f standard.
+    ///
+    /// **Requirements**:
+    /// - All requirements of PDF/A-4
+    A4F,
+    /// The validator for the PDF/A-4e standard.
+    ///
+    /// **Requirements**:
+    /// - All requirements of PDF/A-4
+    A4E,
+}
+
+/// The universal accessibility standard to export the PDF to.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+pub enum UniversalAccessibility {
     /// The validator for the PDF/UA-1 standard.
     ///
     /// **Requirements**:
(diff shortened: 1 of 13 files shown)


crates/krilla/src/configure/validate.rs
```diff
@@ -292,28 +308,155 @@ pub enum Validator {
...
+        }
+    }
+
+    /// Create a validator set from a number of standards.
+    ///
+    /// Returns `None` if the selected validators are not mutually compatible.
+    pub fn with(
+        archival: Option<Archival>,
+        universal_accessibility: Option<UniversalAccessibility>,
+    ) -> Option<Self> {
```
laurmaedje, line 337: API-design wise this feels not super extensible to future potential standards like PDF/X.

laurmaedje (commented):
Not necessarily arguing against it, but I'm curious what you'd say the main benefit of splitting the types up into archival and accessibility standards is.

LaurenzV:
Mainly, I just think that passing a vector of validators (like is done in the other PR) is not the best from an API perspective. Sure, all the constraints can easily be validated at runtime, but the user can still “accidentally” do things like 1) passing two invalid validators like A1 and A2 at the same time and 2) passing the same validator multiple times. Obviously there is no sane reason why someone would do that, but since there seemingly is a way of representing this in a static way in the API, Inthink this would be preferable. 😄 But happy to discuss more if there are concerns!

crates/krilla/src/configure/validate.rs
LaurenzV: True, but the cadence of PDF releases is slow enough that I don’t think it should be a problem to make a breaking release if we add a new one. 😄

crates/krilla/src/configure/validate.rs
```diff
@@ -221,6 +216,27 @@ pub enum Validator {
     /// **Requirements**:
     /// - All requirements of PDF/A-2b
     A3_U,
+    /// The validator for the PDF/A-4 standard.
+    ///
+    /// **Requirements**:
+    /// - While not required, it's recommended to enable tagging.
```
reknih, line 222: ik that this just appears added because of the move, but the requirements section (also of the validators above) seems to devolve into incompleteness. I'd drop it

crates/krilla/src/configure/validate.rs
```diff
@@ -292,28 +308,155 @@ pub enum Validator {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Validators {
+    archival: Option<Archival>,
+    universal_accessibility: Option<UniversalAccessibility>,
+    recommended_version: PdfVersion,
+}
+
+impl Validators {
+    /// Create a [`Validators`] object without any active validators.
+    pub fn dummy() -> Self {
```
reknih, line 323: Isn't this just `Validators::default`? `dummy` seems pejorative

crates/krilla/src/configure/validate.rs
reknih: yes, but we'd also break this if we e.g. would add support for PDF/X, PDF/VT

crates/krilla/src/configure/validate.rs
```diff
@@ -292,28 +308,155 @@ pub enum Validator {
...
+        if min_version > max_version {
+            return None;
+        }
+
+        // If possible, we always prefer 1.7 since it represents the best balance
+        // between not too old and not too new (for now).
+        // Otherwise, just take the maximum possible version.
+        if min_version <= PdfVersion::Pdf17 && PdfVersion::Pdf17 <= max_version {
+            Some(PdfVersion::Pdf17)
+        } else {
```
reknih, line 417: Effectively, this only applies to the naked PDF export, which the comment could explain

crates/krilla/src/configure/validate.rs
```diff
@@ -510,295 +653,309 @@ impl Validator {
...
+    }
+
+    pub(crate) fn write_xmp(self, xmp: &mut XmpWriter, all_validators: &Validators) {
+        use Archival::*;
+        use UniversalAccessibility::*;
+
         if self.is_pdf_a() {
+            let combined_with_ua1 = all_validators
+                .iter()
+                .any(|validator| validator == Validator::UniversalAccessibility(UA1));
```
reknih, line 712: Effectively, this is needed for all UA parts, so you could check whether the `Validators::universal_accessibility` is some

reknih (commented):
I think that this alternative design does not address the weakest link in my PR: The inline `.any` / `.all` merging of policies.

Instead, this implementation aims to preclude the user from choosing illegal combinations of validators by segregating them by ISO standard. I am not convinced that the additional struct and the greater potential for downstream breakage Laurenz raised is justified by the benefits:

- This design prevents the user from setting multiple validators of different parts of the same standard. That's a good idea, but there are more reasons for standards to be mutually incompatible.
   Standards can have version ranges without an overlap (e.g the combination PDF/A-2a and PDF/UA-2 is illegal because they share no allowed versions). They can also declare directly opposing policies. For example, PDF/X-4p requires an external ICC profile while all PDF/A standards prohibit external files.
   I think that a design seeking to improve the experience of remedying the use of mutually incompatible profiles would return a `Result<ConfigurationError>` when constructing `Configuration` where configuration error is an enum like this:
   ```rs
   enum ConfigurationError {
     ValidatorPdfVersionMismatch(Validator, Range<PdfVersion>, Validator, Range<PdfVersion>),
     ValidatorSameFamilyDifferentPart(Validator, Validator),
     ValidatorIncompatiblePolicy(Validator, Validator),
   }
   ```
- I think that downstream consumers will rarely split out standards by family. Typst has the `--pdf-standards` flag with a comma-separated list, and I think other apps will offer a similar experience, so they will need to convert from a flat list to a typed list.

I think the two most worthwhile improvements over my original design would be to eagerly collect the Validators into a ValidationPolicy. This validation would be the one single place to DRYly define the folding behavior of properties, removing duplicate implementations like we both have for embedded / associated files. The other improvement over my design I'd suggest would be to return a Result when constructing `Configuration` instead of an option.

Yet, thank you for so responsively taking care of this, I'm sorry to deliver a rather critical review!

Addendum: We could also make the validators a `BTreeSet`. Since the set will in practice at most have four entries (PDF/A, PDF/UA, PDF/X, PDF/VT), this will be slower than a vector, but it would remove the failure case of adding the same validator multiple times.

LaurenzV:
Hmm fair enough, thanks for taking a look! I think having a single `ValidationPolicy` struct would indeed be the best, but I'm fine just going with what you have in your PR for now. Can't promise when I'll be able to review that, so feel free to merge if it looks good to either @laurmaedje or @saecki! 🙂

saecki:
> I think that this alternative design does not address the weakest link in my PR: The inline `.any` / `.all` merging of policies.
> 
> Instead, this implementation aims to preclude the user from choosing illegal combinations of validators by segregating them by ISO standard. I am not convinced that the additional struct and the greater potential for downstream breakage Laurenz raised is justified by the benefits:
> * This design prevents the user from setting multiple validators of different parts of the same standard. That's a good idea, but there are more reasons for standards to be mutually incompatible.
>   Standards can have version ranges without an overlap (e.g the combination PDF/A-2a and PDF/UA-2 is illegal because they share no allowed versions). They can also declare directly opposing policies. For example, PDF/X-4p requires an external ICC profile while all PDF/A standards prohibit external files.
>   I think that a design seeking to improve the experience of remedying the use of mutually incompatible profiles would return a `Result<ConfigurationError>` when constructing `Configuration` where configuration error is an enum like this:
>   ```rust
>   enum ConfigurationError {
>     ValidatorPdfVersionMismatch(Validator, Range<PdfVersion>, Validator, Range<PdfVersion>),
>     ValidatorSameFamilyDifferentPart(Validator, Validator),
>     ValidatorIncompatiblePolicy(Validator, Validator),
>   }
>   ```
>
> * I think that downstream consumers will rarely split out standards by family. Typst has the `--pdf-standards` flag with a comma-separated list, and I think other apps will offer a similar experience, so they will need to convert from a flat list to a typed list.

I somewhat disagree with this point. I think it's a good thing to split the standards into families, even when standards from different families can have additional non-trivial incompatibilities that are reported by the `ConfigurationError` you outlined above.
Only more recently coming into contact with PDF specs, I think having a strict user-facing separation into families would've made things clearer for me. And I think it would especially make sense to reflect this in the typst web-app UI.
Yes this would be *a little* more work for consumers of `krilla`, but for typst-cli it would only mean one of two things:
- Either change the `--pdf-standard` flag could be split up into `--pdf-version`, `--pdf-a-standard`, `--pdf-ua-standard`. Not sure about that, because it would be quite breaking, but I think that would've been the cleaner design from the start.
- The existing mapping of standards into versions and validators has to be extended to also group into families. I've always found it a bit odd that versions and validators were grouped into one CLI flag, but the additional complexity doesn't seem to bad.

> I think the two most worthwhile improvements over my original design would be to eagerly collect the Validators into a ValidationPolicy. This validation would be the one single place to DRYly define the folding behavior of properties, removing duplicate implementations like we both have for embedded / associated files.

I think a resolved `ValidationPolicy` would be great, but is a somewhat orthogonal concern to the consumer-facing API of how the configuration is constructed. The split into families should happen either way, either internally of externally.
I don't think there is a duplication in the folding behavior. Separating into families is the first step, finding incompatibilities between standards of different families it the second one.

> The other improvement over my design I'd suggest would be to return a Result when constructing `Configuration` instead of an option.

This would be nice indeed.

PR #358, opened 2026-04-23
Support some more composite modes in COLR

LaurenzV:
Apart from standard blend modes like Multiply, Screen, Darken, etc., the COLR imaging model also allows for using different Porter-Duff compositing operators, which PDF does not support natively.

However, the key insight is that some of those blend modes can be simulated by rendering either the source/backdrop twice, once normally and once into a mask, and then use that mask for the compositing operation. For example, `SrcIn` basically consists of not drawing the backdrop at all, but instead drawing it into a mask and then drawing the source with that mask:
image

After looking at all prominent COLR fonts (Noto Color Emoji and Segoe UI Emoji), it turns out that SrcIn is actually the _only_ other Porter Duff composition operator that is used. As a consequence, while this PR doesn't implement all of them (as some are a bit trickier), we can now render all emoji glyphs correctly. See also the before vs. after after hooking this change into Typst:

[before.pdf](https://github.com/user-attachments/files/27027991/before.pdf)
[after.pdf](https://github.com/user-attachments/files/27027992/after.pdf)


refs/visreg/font_colr_test_glyphs.png
LaurenzV, line 1: The changing colors for the crosses is expected, compare to https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/web_tests/platform/linux/virtual/text-antialias/colrv1-expected.png. Looks like I somehow got that wrong before, I'm not entirely sure what caused this, though.

refs/visreg/font_noto_color_emoji_colr_ghostscript.png
LaurenzV, line 1: Similarly, the change in direction of the tea can also is expected, when I compare it to PNG output in Typst.

laurmaedje:
Nice!

PR #382, opened 2026-05-07
Add invisible glyph text layer helper

hongjr03:
This adds a helper for writing a glyph run with PDF text rendering mode 3 (`Tr 3`), defined in ISO 32000-2 section 9.3.6 as “Neither fill nor stroke text (invisible)”.

That lets callers keep a text layer for copy/paste, search, ToUnicode mappings, and tagging even when the visible glyphs are drawn separately.

### DIFF
--- a/crates/krilla/src/content.rs
+++ b/crates/krilla/src/content.rs
@@ -487,6 +487,36 @@ impl ContentBuilder {
         self.graphics_states.restore_state();
     }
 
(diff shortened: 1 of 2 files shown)


LaurenzV:
Couldn’t you just set the fill opacity to 0?

hongjr03:
Sorry I didn't notice that.

LaurenzV:
All good!

Issue #313, opened 2025-11-29
Marked content inside a form XObject?


mwcampbell:
I believe I've encountered a limitation of either Krilla or PDF itself. To reproduce, see [this version of the parley example](https://gist.github.com/mwcampbell/af79397d3f80df9cd100154a3094932a), where I'm both creating a tagged PDF and wrapping all drawing operations inside a form XObject (specifically an isolated transparency group). With this combination, neither Poppler's `pdfinfo -struct-text` nor Acrobat can locate the marked content. So I wonder if it's even possible to have marked content inside a form.

PR #386, opened 2026-05-15
Always use correct page root transform for artifact bbox

saecki:
Depends on #385

As mentioned in https://github.com/LaurenzV/krilla/pull/273#discussion_r3246334591, any `push` operation on the surface that would cause a builder to be inserted into the `sub_builders` array would use the identity `root_tranform` of the sub builder instead of the page root transform inside the `root_builder`.

(No idea why the macos CI is failing)

### DIFF
--- a/crates/krilla-tests/src/tagging.rs
+++ b/crates/krilla-tests/src/tagging.rs
@@ -5,6 +5,7 @@ use krilla::annotation::{LinkAnnotation, Target};
 use krilla::error::KrillaError;
 use krilla::geom::{PathBuilder, Point, Rect, Size, Transform};
 use krilla::metadata::Metadata;
+use krilla::num::NormalizedF32;
 use krilla::outline::Outline;
 use krilla::page::PageSettings;
 use krilla::paint::{Fill, Stroke};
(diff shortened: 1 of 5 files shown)


LaurenzV:
Is it possible to add a snapshot test that shows that the previous behavior is wrong? DOn't worry about the MacOS thing, I'll fix it later

LaurenzV:
I might remember wrong, but I think we actually don't support tagging in XObjects. 🤔 So it's possible that this actually cannot be triggered, but I'm not sure right now.

saecki:
I've added a test that produces a different `/Bbox` value, you can see the change in diff of the second commit. I'm not sure the produced PDF is valid though :thinking:

LaurenzV:
I'll try to check in the evening.

LaurenzV:
Why do you think the produced PDF is not valid?

LaurenzV:
There _is_ the problem that something like this will lead to a wrongly-tagged PDF:

```
#[snapshot(document)]
fn tagging_span_in_xobject(document: &mut Document) {
    let mut tag_tree = TagTree::new();
    let mut par = TagGroup::new(Tag::P);

    let mut page = document.start_page();
    let mut surface = page.surface();

    surface.push_opacity(NormalizedF32::new(0.5).unwrap());
    let id = surface.start_tagged(ContentTag::Span(SpanTag::empty()));
    surface.fill_text_(100.0, "tagged inside xobject");
    surface.end_tagged();
    surface.pop();

    surface.finish();
    page.finish();

    par.push(id);
    tag_tree.push(par);

    document.set_tag_tree(tag_tree);
}
```

which kind of relates to #337 and #313. I assumed it's not possible in the krilla API to add tags to XObjects, but I forgot that XObjects can be created transparently via methods like `push_opacity`. That definitely needs to be fixed! Though I believe this doesn't affect Typst because we don't make use of those.

PR #389, opened 2026-05-18
Always add `/Tabs S` for PDF/UA and PDF/A-*a

reknih:
Addresses typst/typst#7990 once merged


LaurenzV:
I'm curious, have you tried what happens when you validate a PDF 1.4 file with PDF/UA in Acrobat? Will it still complain then? 🤔

LaurenzV:
Would be nice to try out in case you have Acrobat, but not blocking. Thanks!

PR #390, opened 2026-05-22
Don't enable the `zlib-rs` feature in flate2

danbishop-canva:
This changes krilla’s flate2 dependency wiring so the zlib-rs backend is exposed through an explicit zlib-rs crate feature instead of being hard-coded directly in the workspace dependency.

zlib-rs remains enabled by default for krilla, preserving the current default behavior, but downstream users can now opt out with default-features = false if they need flate2 to resolve to its Rust/miniz backend instead.

Also adds a matching zlib-rs feature to krilla-svg so SVGZ decoding and krilla’s own compressed stream handling can be configured consistently.


LaurenzV:
Thanks! I've decided to not make this a feature in krilla and instead just use the default backend. If others prefer using zlib-rs, they can just add the feature themselves by adding a direct dependency on flate2 and feature unification will take care of the rest.

PR #393, opened 2026-05-26
Add full font bbox option to serialize setting 

winterSteve25:
Adds an option to `SerializeSetting` to use the full font bbox instead of computing the bbox for the subset font and only fallback to the full font bbox.


LaurenzV:
Hmm, sorry, but I don't really see the point in having this, could you please elaborate on your use case?

winterSteve25:
I am working on a DOCX to PDF converter that needs to produce structurally similar PDFs to Microsoft Word. Although your approach is probably better, Microsoft Word uses the full bbox. I feel like if we want this behaviour, there are probably others that do as well. If this seems like an unnecessary change, I understand. We will get by with a crate patch.

LaurenzV:
Okay, I understand, thanks for providing the context! 

But yeah, I'm afraid I'm not sure I would want to land this. :( For a few reasons:
1) This will cause issues because the subsetted font still has the new bbox, so the PDF metadata and font metadata will diverge (which IIRC is for example prohibited by PDF-A).
2) The bbox might be inaccurate for instanced variable fonts (I believe by default the global bbox of a font does not take variation axes into account).
3) While I definitely want to accommodate as many use cases as possible, to me this still seems veeery niche, and I'm a bit wary of adding too many knobs to the already very big `SerializeSettings` struct. 😄 

I hope you understand! I might still reconsider this depending on whether there really are others that would like this, but in that case I would implement it properly such that the subsetted font also uses the original bbox.

winterSteve25:
All good :) I understand, thanks for your time.

PR #398, opened 2026-05-29
Don't unconditionally reject header/footer artifacts

### LINKED ISSUE #8383: Page headers no longer work at all in accessible documents after recent Krilla update
### Description

The document
```typst
#set document(
  title: "Bug with accessible headers"
)

#set page(
  header: [aaa]
)

#title()

#lorem(20)
```
produces the following error when using PDF/UA-1 as the output format:
```sh
error: PDF/UA-1 error: headers and footers cannot be made accessible in PDF 1.7 files
 = hint: set the version to PDF 1.7
```
Note the that I am already using PDF version 1.7 as the output target so the hint is useless. To my knownledge, previously headers were treated as artefacts, but something apparently broke with <https://github.com/typst/typst/pull/8294> .

### Operating system

macOS

### Typst version

- [x] I am using the latest version of Typst

LaurenzV:
We already had a version check for two of the artifacts, but forgot it for headers/footers.

Verified that this fixes https://github.com/typst/typst/issues/8383.

### DIFF
--- a/crates/krilla-tests/src/validate.rs
+++ b/crates/krilla-tests/src/validate.rs
@@ -2,7 +2,7 @@ use krilla::action::LinkAction;
 use krilla::annotation::{Annotation, LinkAnnotation, Target};
 use krilla::color::{rgb, separation};
 use krilla::configure::validate::VersionedFeature;
-use krilla::configure::ValidationError;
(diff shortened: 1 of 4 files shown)


reknih (approved):
LGTM. Yeah, sorry for the mistake

LaurenzV:
We both missed it! 😄

PR #401, opened 2026-06-03
Soften requirement for Artifact bboxes

reknih:
Krilla previously asked for bounding boxes for all background artifacts, independent of PDF version. This type was introduced in PDF 1.7, but its mandate to come with BBoxes has been lifted by PDF 2.0.

Confer the Adobe-published edition of PDF 32000-1:2008, Table 330, and ISO 32000-2:2020(E), Table 363.

In practice, at least in Typst, the bounding box is not readily available for some background artifacts. Instead of writing the whole page to write something, we'd rather write nothing in PDF 2.0 in these cases.

If this is not the right spot to panic, happy to move or remove this check altogether.

### DIFF
--- a/crates/krilla/src/interchange/tagging/mod.rs
+++ b/crates/krilla/src/interchange/tagging/mod.rs
@@ -162,13 +162,10 @@ pub struct Artifact {
 impl Artifact {
     /// Create a new artifact with a type and an optional BBox.
     ///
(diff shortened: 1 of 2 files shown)


saecki (approved):

LaurenzV:
Looks good, we probably want to just have an error for this in the future, but this will require rethinking the current error architecture a bit.

PR #402, opened 2026-06-03
Support variable fonts in `krilla-svg`

laurmaedje:
This PR adds support for variable fonts to `krilla-svg`.

To arrive at this, I first refactored the font handling in `krilla-svg` a bit (https://github.com/LaurenzV/krilla/commit/51f08e70e1e14b763830e5f0ee4a3964086a87ab). Previously, I had tried a [minimally invasive patch](https://github.com/LaurenzV/krilla/commit/1d549ba44bc4441d636eae68e8ccbc6cf5460653), but supporting optical sizing and filtering out undefined axes got a bit messy. The refactoring in the first commit of this PR simplifies the font handling quite a lot by removing the pre-pass and just lazily constructing fonts as we go. I think this change _should_ be safe, but if there's a reason for the existing setup, please let me know!

The main change is just a bit of plumbing to forward the font variations from usvg to krilla's `Font`. However, a bit of care must taken in the process. I first just forwarded the font variations unconditionally, but that failed some tests because resvg will happily specify font variations even if the font isn't variable. [^1]

To avoid this, I've implemented logic that retains only the variations that are actually defined in the font before passing them on to the krilla `Font`.

I also implemented support for `span.font_optical_sizing` based on how it's implemented in `resvg`. I can confirm that with these changes, Typst's PDF and PNG export for an SVG containing text with Roboto Flex (which supports optical sizing) match.

I've added a snapshot test with a very basic variable font test (maybe a visreg test would preferrable?). More tests would probably be good, but it's better than nothing.

[^1]: I'm not entirely sure why this failed the tests tbh. The reference images [changed](https://github.com/LaurenzV/krilla/actions/runs/26902471613/job/79358255399) ever so slightly ...


LaurenzV:
image
Slant one looks a bit sus, that's probably what your subsetter PR should fix? Otherwise looks nice!

laurmaedje:
> Slant one looks a bit sus, that's probably what your subsetter PR should fix?

Probably! Not sure how you generated that image, but if you could try with the patch that would be nice.

laurmaedje:
If that fixes it, I would go ahead, merge the subsetter PR, and cut a subetter patch release.

LaurenzV:
Yup, one sec

LaurenzV:
image
Yup works!

laurmaedje:
Just published subsetter 0.2.5.

LaurenzV:
I'm trying to remember why I chose this preprocessing approach. I'm pretty sure it was a conscious decision, but I can't remember right now. 😢

LaurenzV:
Okay, I let codex do some digging and I think this rings a bell. Basically, font handling worked much different back then before the current `Font` structure was introduced that is prehashed and also cached the `FontRef`, which required this hack to ensure that multiple fonts are handled efficiently when embedding multiple SVGs in a single document. I'll think about it again as well, but the change should be safe!

image

laurmaedje:
Yeah, my suspicion was also that it's a historical artifact.

LaurenzV:
Thank you!!

PR #405, opened 2026-06-04
Tighten visibilities in `krilla-svg`

LaurenzV:


laurmaedje:
Curious about this. Just personal preference? Since the `ProcessContext` is already private it doesn't make a difference, right?

LaurenzV:
For ProcessContext yeah, but for fonts it makes easier to see when just looking at the code without how everything is exported that changing those isn't a public breaking change 😄

PR #408, opened 2026-06-16
Support writing custom XMP metadata

### LINKED ISSUE #209: Support custom XMP metadata in PDFs
Support writing custom XMP metadata for things like e-invoices or writing RDF license information.

All XMP metadata for Typst's PDF export is written inside of krilla as far as I can see. That means typst/typst#5667 will require changes in krilla.

NiklasEi:
This PR allows writing custom XMP metadata. It checks for predefined schemas for PDF/A-1 to A-3, makes sure we don't conflict with prefixes known by `xmp-writer`, validates all custom namespaces to be consistent and checks for duplicated prefixes.
It does not support defining custom types. But as the example/test shows, you can define e-invoices following the ZUGFeRD standard with this, which is my main use-case.

For now, the custom properties have a single setter following the example of the other metadata parts. The namespaces are duplicated in the properties (I think I'll try a builder pattern though; I don't really like the current approach).

Resolves #209


crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +620,180 @@ impl PageLayout {
...
+
+/// Property schemas added by the XMP 2005 specification over the XMP 2004 set
+/// in [`PREDEFINED_PDFA_SCHEMAS_XMP_2004`]. PDF/A-2 and
+/// PDF/A-3 reference XMP 2005 rather than restating a list.
+///
+/// The three schemas are defined by the Adobe XMP Specification, June 2005:
+/// Dynamic Media (`xmpDM`), Camera Raw (`crs`) and EXIF auxiliary (`aux`).
+/// They are the schemas present in that edition but absent
+/// from XMP 2004.
+static PREDEFINED_PDFA_SCHEMAS_XMP_2005_ADDITIONS: &[&str] = &[
```
NiklasEi: The TN 0008 with the list of predefined schemas above was easy to find. But for these three here I ended up trusting veraPDF.

tests/src/metadata.rs
```diff
@@ -56,3 +62,370 @@ fn metadata_pdf_20_author(document: &mut Document) {
...
+    metadata_impl(document);
+
+    let cc = cc_namespace();
+    let mm = Namespace::new("xmpMM", "http://ns.adobe.com/xap/1.0/mm/");
+
+    let metadata = Metadata::new()
+        .creation_date(datetime())
+        .custom_xmp_properties(vec![
+            Property::new(
+                cc.clone(),
```
NiklasEi, line 108: Alternatively, we could create the property on the namespace itself.

tests/src/metadata.rs
```diff
@@ -56,3 +62,370 @@ fn metadata_pdf_20_author(document: &mut Document) {
...
+            Property::new(fx.clone(), "DocumentType", Value::text("INVOICE")),
+            Property::new(fx.clone(), "DocumentFileName", Value::text("factur-x.xml")),
+            Property::new(fx.clone(), "Version", Value::text("1.0")),
+            Property::new(fx, "ConformanceLevel", Value::text("BASIC")),
+        ]);
+    document.set_metadata(metadata);
+}
+
+/// A Factur-X invoice in the MINIMUM profile.
+const FACTUR_X_MINIMUM_XML: &str = r#"<?xml version="1.0" encoding="UTF-8"?>
```
NiklasEi, line 178: This is not a fully fledged invoice, but I ran the PDF through a validator and the metadata seems fine.

LaurenzV:
Thanks, will take a look when I find some time! Perhaps someone from the Typst folks can also take a look.

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +620,180 @@ impl PageLayout {
         }
     }
 }
+
+/// The namespaces xmp-writer can serialize natively. A user-supplied namespace
+/// with one of these URLs must be mapped to the corresponding built-in variant,
+/// since xmp-writer would otherwise declare the same namespace twice.
```
NiklasEi, line 623: I think these should not always get filtered out of the custom namespaces to write extension schemas for. We should just make sure that the prefixes/URIs from `xmp-writer` get used. Some of these are not predefined for PDF/A, so we should write an extension schema.

NiklasEi: While I was writing the test for this, I noticed a difference between the namespace URIs in xmp-writer and Vera PDF -> typst/xmp-writer#6

LaurenzV:
(Haven't forgotten about this, but unfortunately I've been away this week and unsure how much time I'll have next one!)

LaurenzV:
@NiklasEi Sorry again for the lack of response, unfortunately haven't had the time yet. 😅 However, the good news is that there is interest from other parties in this as well, so we will land this in some form soon. Would it be okay with you if @danbishop-canva does some more work/fixes based on your branch, I'll take a brief look afterwards and then we can merge it?

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -1,17 +1,23 @@
 //! Setting document metadata.
 //!
 //! PDF allows for the inclusion of metadata in a PDF document. To do so in krilla,
-//! you can simply create a [`Metadata`] object, set the data, and then include it
+//! you create a [`Metadata`] object, set the data, and then include it
```
reknih: Drop the `you`, it's cleaner.
(So "To do so in krilla, create a ...")

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
         }
     }
 }
+
+/// The namespaces xmp-writer can serialize natively. A user-supplied namespace
+/// with one of these URIs must be mapped to the corresponding built-in variant,
+/// since xmp-writer would otherwise declare the same namespace twice.
+static BUILTIN_XMP_NAMESPACES: &[XmpNamespace<'static>] = &[
```
reknih: Have you considered exposing this list from `xmp-writer` itself instead, e.g. as an iterator method on the [`Namespace`](https://docs.rs/xmp-writer/latest/xmp_writer/enum.Namespace.html) enum? Otherwise, this invites drift on updates.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Namespace {
+    /// The XML prefix (e.g. `"fx"`).
+    pub prefix: String,
+    /// The namespace URI (e.g. `"urn:factur-x:pdfa:CrossIndustryDocument:invoice:1p0#"`).
+    pub uri: String,
+    /// Optional human-readable schema name for the PDF/A extension schema
+    /// description. Defaults to `"<prefix> schema"`.
+    pub schema_name: Option<String>,
+    /// PDF/A extension-schema property descriptions.
```
reknih: The spec spells `extension schema` without a hyphen. Adjust throughout.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub name: String,
+    /// The value type. Either a built-in XMP type (`"Text"`, `"Integer"`,
+    /// `"Date"`, ...) or a custom value type.
+    pub value_type: String,
+    /// Whether the property is generated internally by the producer or
+    /// supplied externally by the user.
+    pub category: Category,
+    /// Human-readable description of the property.
+    pub description: String,
+}
```
reknih, line 111: This approach does not allow to write RELAX NG schemata for PDF/A-4. In particular, `value_type` is not machine-readable. It is also unable to model the `required`ness of a field.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// An XMP property value.
+///
+/// The variants mirror the value shapes supported by XMP/RDF: primitive
+/// types, the three array kinds (`rdf:Seq`, `rdf:Bag`, `rdf:Alt`), language
+/// alternatives, and structs.
+#[derive(Debug, Clone, PartialEq)]
+pub enum Value {
+    /// A text value.
+    Text(String),
```
reknih, line 309: Both this API and `xmp-writer` currently do not handle URLs correctly. Instead of being serialized as an XML text node, they need to be placed in an `rdf:resource` attribute of the otherwise childless parent node (confer XMP Specification Part 1, Note in Section 6.3.2, Section 7.5; [RDF 1.1 XML Syntax Section 2.4](https://www.w3.org/TR/rdf-syntax-grammar/#h3_section-Syntax-empty-property-elements))

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+impl StructField {
+    /// Create a new struct field.
+    pub fn new(namespace: Namespace, name: impl Into<String>, value: Value) -> Self {
+        Self {
+            namespace,
+            name: name.into(),
+            value,
+        }
+    }
+}
```
reknih: This struct is identical to `Property`

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// Convenience constructor for a [`Value::Text`].
+    pub fn text(value: impl Into<String>) -> Self {
+        Self::Text(value.into())
+    }
+
+    /// Convenience constructor for a [`Value::Date`].
+    pub fn date(value: DateTime) -> Self {
+        Self::Date(value)
+    }
+}
```
reknih, line 343: Should we provide `From` implementations for these types? What about the numeric types?

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+pub enum Category {
+    /// The property is computed by the producer (e.g. page count).
+    Internal,
+    /// The property is supplied by the user.
+    External,
+}
+
+/// A single XMP property attached to a [`Namespace`].
+#[derive(Debug, Clone, PartialEq)]
+pub struct Property {
```
reknih, line 281: This is unable to model qualifiers.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+//! Custom namespaces must not reuse the prefix of a predefined schema or
+//! bind one prefix to two different URIs, and all declarations of the same
+//! URI must be identical; otherwise [`Document::finish`] returns an
+//! [`XmpError`].
+//!
+//! ## PDF/A
+//!
+//! Any property in a custom namespace must be described in a PDF/A
+//! extension schema. Populate [`Namespace::property_descriptions`] for
+//! every property you write.
```
reknih: I think that requiring extension schema data for non PDF/A use cases is too strict. Consider a [TDMRep](https://www.w3.org/community/reports/tdmrep/CG-FINAL-tdmrep-20240510/#sec-pdf) assertion in XMP: They are expected to occur in non-PDF/A files and this requirement would otherwise force a user to specify something we discard later.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// An XMP namespace.
+///
+/// A namespace is identified by its URI. The prefix is the short name used
+/// in the serialized XML. For PDF/A output, the
+/// namespace must describe each property through
+/// [`Self::property_descriptions`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Namespace {
+    /// The XML prefix (e.g. `"fx"`).
```
reknih, line 44: Document that this may not be used for built-in schemata.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// A [`Value::Real`] was NaN or infinite, which can't be represented in
+    /// XMP. Contains the name of the containing property.
+    NonFiniteReal(String),
+}
+
+impl std::fmt::Display for XmpError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            XmpError::ConflictingNamespace(uri) => {
+                write!(f, "the namespace {uri} was declared inconsistently")
```
reknih:
```suggestion
                write!(f, "the namespace {uri} was declared with multiple different prefixes")
```

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
...
+    BUILTIN_XMP_NAMESPACES
+        .iter()
+        .find(|ns| ns.url() == uri)
+        .cloned()
+}
+
+fn is_reserved_xmp_prefix(prefix: &str) -> bool {
+    BUILTIN_XMP_NAMESPACES
+        .iter()
+        .any(|ns| ns.prefix() == prefix)
```
reknih: On a statically present list, binary search could be possible.

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
...
+}
+
+/// Predefined XMP property schemas a custom property may use in PDF/A-1
```
reknih: What about PDF/A-2 and A-3?

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -242,6 +256,131 @@ impl Metadata {
...
+                    .namespace
+                    .property_descriptions
+                    .iter()
+                    .any(|desc| desc.name == property.name)
+            {
+                sc.register_validation_error(ValidationError::MissingXmpPropertyDescription {
+                    namespace_uri: property.namespace.uri.clone(),
+                    property_name: property.name.clone(),
+                });
+            }
```
reknih, line 290: The type, e.g. the value of a closed choice is not validated throughout. Worth discussing whether this is the responsibility of the caller or Krilla, but Krilla performs similar checks elsewhere.

crates/krilla/src/interchange/metadata/mod.rs
reknih: Does the standard back this, either in a clause or by way of a different citation in its bibliography?

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
...
+}
+
+pub(crate) fn write_user_extension_schema(
+    schemas: &mut xmp_writer::pdfa::PdfAExtSchemasWriter,
+    ns: &Namespace,
+) {
+    let xmp_namespace = build_xmp_namespace(ns);
+    // A namespace xmp-writer knows natively is serialized under its canonical
+    // prefix, not the user-supplied one.
+    let prefix = xmp_namespace.prefix();
```
reknih, line 757: xmp-writer has its own mechanism to describe schemas not built in. Do you prevent a collision if, e.g. the user uses a property from the `http://www.npes.org/pdfx/ns/id/` schema?

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
...
+        .value(prefix);
+
+    if !ns.property_descriptions.is_empty() {
+        let mut properties = schema.properties();
+        for desc in &ns.property_descriptions {
+            properties
+                .add_property()
+                .name(desc.name.as_str())
+                .value_type(desc.value_type.as_str())
+                .category(matches!(desc.category, Category::Internal))
```
reknih: `Category` implements `Eq`:
```suggestion
                .category(desc.category == Category::Internal)
```

crates/krilla/src/configure/validate.rs
```diff
@@ -142,6 +144,14 @@ pub enum ValidationError {
     EmbeddedPDF(Option<Location>),
     /// A feature only available in a later PDF version was required.
     RequiresNewerPdfVersion(VersionedFeature, Option<Location>),
+    /// A custom XMP property was used in PDF/A export without a matching
+    /// PDF/A extension schema description.
+    MissingXmpPropertyDescription {
```
reknih, line 149: No additional error for forbidden properties, see [TechNote 0008](https://pdfa.org/wp-content/uploads/2011/08/tn0008_predefined_xmp_properties_in_pdfa-1_2008-03-20.pdf)

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -242,6 +256,131 @@ impl Metadata {
...
+    ) {
+        if self.custom_xmp_properties.is_empty() {
+            return;
+        }
+
+        let validators = sc.serialize_settings().validators();
+        let check_descriptions = validators.requires_xmp_metadata_extension_schema();
+        let xmp_2005 = validators.uses_xmp_2005_predefined_schemas();
+
+        for property in &self.custom_xmp_properties {
```
reknih, line 277: You are iterating over this vector at least three distinct times. Can you write an implementation that simulataneously writes, validates, and collects namespaces?

NiklasEi:
> Would it be okay with you if @danbishop-canva does some more work/fixes based on your branch, I'll take a brief look afterwards and then we can merge it?

Of course :slightly_smiling_face:

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+    Internal,
+    /// The property is supplied by the user.
+    External,
+}
+
+/// A single XMP property attached to a [`Namespace`].
+#[derive(Debug, Clone, PartialEq)]
+pub struct Property {
+    /// The namespace this property belongs to.
+    pub namespace: Namespace,
```
reknih, line 283: `Namespace` is a pretty heavy struct that is reused across multiple properties and later deduplicated. This kind of pattern exists in other parts of Krilla but is typically handled via reference counting and sometimes pre-computing a hash. Worth thinking about.

crates/krilla/src/interchange/metadata/xmp.rs
reknih: Also, for RELAX NG, a `pub description: Option<String>` is required.

crates/krilla/src/interchange/metadata/xmp.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+/// Description of a single XMP property under a [`Namespace`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PropertyDescription {
+    /// The property name (e.g. `"DocumentType"`).
+    pub name: String,
+    /// The value type. Either a built-in XMP type (`"Text"`, `"Integer"`,
+    /// `"Date"`, ...) or a custom value type.
```
reknih: Where are the custom value types defined?

crates/krilla/src/interchange/metadata/mod.rs
```diff
@@ -475,3 +614,185 @@ impl PageLayout {
...
+
+/// Predefined XMP property schemas a custom property may use in PDF/A-1
+/// without a PDF/A extension schema description (PDF/A-1 is based on XMP 2004).
+/// Also the common base of the PDF/A-2/-3 set.
+///
+/// These are the standard schemas of the XMP 2004 specification that ISO
+/// 19005-1 (PDF/A-1) treats as predefined. The PDF Association restates the
+/// list in TechNote 0008, "Predefined XMP Properties in PDF/A-1":
+/// <https://pdfa.org/wp-content/uploads/2011/08/tn0008_predefined_xmp_properties_in_pdfa-1_2008-03-20.pdf>
+static PREDEFINED_PDFA_SCHEMAS_XMP_2004: &[&str] = &[
```
reknih: You cannot treat whole namespaces as predefined. Instead, this needs to be allow-listed on a per-property bases. This bit us in the past, especially with `xmpMM:InstanceID`.

reknih (commented):
Overall, I think that this PR does not solve the entire problem of custom XMP. In particular, some of the missing parts would be hard to retrofit. We have two distinct problems:

1. Merging the user-provided XMP properties with the properties Krilla manages.
2. Self-describing the schema in PDF/A: Inline with extension schemas in part 1 through 3 and in an associated RELAX NG file in part 4.

With the second problem in particular, this PR does not yet provide the machinery neccessary for a mature integration.
In principle, the proposed API (a `Namespace` struct with a vector of `PropertyDescription`s) is closely aligned with PDF/A extension schemas. Yet, it is unable to represent structure types due to a missing field for value types on the `Namespace` struct. However, the larger problem is that it inherits the weakness of the extension schemas at large: `PropertyDescription::value_type` is an opaque string, matching its definition in ISO 19005-2:2011, Section 6.6.2.3.3, as an "Open Choice of Text". This means that Krilla cannot validate the user-provided XMP against the schema and it cannot generate a RELAX NG schema for PDF/A-4 (c.f. ISO 19005-4:2020 Section 6.7.2.3, ISO 16684-2).

Parsing this string to generate RELAX NG is not feasible since no standard defines a strict-enough grammar beyond "Value type of the property, drawn from XMP Specification, or an embedded PDF/A extension schema value type" for this value. In the wild, for example, I have already seen these strings localized.

This means that, without API breakage, Krilla can neither validate that user-provided XMP matches the schema, nor generate RELAX NG for PDF/A-4 files (which is a "should" recommendation). Instead, I propose that Krilla adopts the XMP type system so it can both validate it and translate to RELAX NG. `value_type` should be able to express at least the simple-valued types in Section 8.2 of Part 1 of the XMP Specification, including simple-valued derived value types as well as the _Rational_ type from Part 2 of the XMP Specification, Section 1.2.6.9. It should also be able to represent the generic types ordered arrays, unordered arrays, alternative lists (XMP Spec Part 1, Section 6.3.4), and language alternatives (ibid., Section 8.2.2.4) with their type parameter and accept references to custom structure types. The API may or may not also support the structure types defined in Section 8.2.2 of the first part and Section 1.2.2 of the second part of the XMP Specification.

This is how we are modelling Typst's API and also how we plan to consume Krilla. Krilla, so far, has had the responsibility of performing format-specific conversions and we think that it is unreasonable to expect Typst or other library consumers to change this for XMP. Furthermore, having Krilla emit some managed XMP and some unmanaged, custom XMP makes it impossible for the consumer to provide a full RELAX NG schema for the totality of metadata at all.

With regards for the semantics of merging Krilla- and user-defined properties, there are few implementation-specific but fixable problems: This PR currently does not raise an error if the user provides a value Krilla already wrote as XMP. I think that the reasonable behavior would be to raise an error on a duplicate or overwrite it without producing an invalid file. Furthermore, writing a custom property in a schema not referenced by TechNote 0008 and yet with some managed schema properties will result in a duplicate schema definition.

crates/krilla/src/interchange/metadata/xmp.rs
NiklasEi: They are not supported in the current draft as mentioned in the MR description. I will either update the documentation here or add support for them.

NiklasEi:
@reknih thank you for the thorough review! I had not considered RELAX NG so far and thought it would need a separate solution anyway, but I see your point. It makes more sense to have one API where we know enough about the values to either write XMP metadata or RELAX NG.

I can reproduce the duplicate schema definition for things Krilla emits as well.

> This PR currently does not raise an error if the user provides a value Krilla already wrote as XMP. I think that the reasonable behavior would be to raise an error on a duplicate or overwrite it without producing an invalid file.

If a user of Krilla explicitly sets a metadata value, we should write it. But I think that only works for the properties that mirror a Metadata field. The ones that get written to satisfy a standard should error. We could keep a whitelist.

crates/krilla/src/interchange/metadata/mod.rs
NiklasEi: typst/xmp-writer#7

crates/krilla/src/interchange/metadata/mod.rs
NiklasEi: Two lines below the docs mention that these are also for A-2/A-3. I restructured them to be clearer and renamed the statics.

crates/krilla/src/interchange/metadata/mod.rs
NiklasEi: TN 0008 mentions them as not allowed in PDF/A-1 because not included in XMP 2004, but introduced in XMP 2005. See footnote 1 in §1.5.

PR #410, opened 2026-06-17
Preserve page group of embedded PDFs

Hyperrick:
## Summary

When a PDF page is embedded as a Form XObject, krilla currently always writes a color space on the outer transparency group. That is useful for PDFs whose page group explicitly defines an RGB color space, but forcing an inferred RGB group can change the composition of print-oriented source PDFs that contain CMYK or spot-color content.

This PR makes the outer XObject group follow the source page instead of guessing from its contents:

- Preserve an explicit `/DeviceRGB` or `/DeviceCMYK` value from the source page's `/Group /CS`.
- Leave `/CS` unspecified when the source page does not explicitly provide one of those group color spaces.
- Do not infer RGB or CMYK from content operations, color-space resources, separations, or `DeviceN` usage.

## Context

I originally tried to detect whether an embedded PDF was "RGB or CMYK" from its content and resources. As discussed in the review, that heuristic is not reliable because a PDF can legitimately mix color spaces and separations.

I retested this with print-production PDFs and compared the result with `pdflatex` and `lualatex`. For source pages without an explicit group color space, their outer embedded XObject group remains without `/CS`, while nested CMYK and separation color spaces stay intact. Matching that behavior also fixed the separation changes I observed in Acrobat Output Preview.

The existing regression fixture from [typst/typst#7269](https://github.com/typst/typst/issues/7269) has an explicit source page `/Group /CS /DeviceRGB`. This PR preserves that explicit RGB value, so the earlier transparency and PDF-viewer compatibility fix remains covered rather than being replaced by a print-specific heuristic.

An analogous color-correctness concern is documented in [OpenPrinting/cups-filters#569](https://github.com/OpenPrinting/cups-filters/issues/569#issuecomment-3707820635): when pages using different `DeviceRGB` and `DeviceCMYK` spaces are wrapped as Form XObjects, assigning a shared group color space can produce incorrect output. That discussion likewise points to avoiding unnecessary group color-space conversions when independently produced artwork is assembled.

## Print workflow impact

Placed PDFs can be preflighted production assets containing process colors, spot colors, or separations. Adding an inferred RGB color space to the outer transparency group can change how those nested color spaces are composed. In practice, that can turn K-only black into process black, shift CMYK values, or affect separations even though the source PDF itself is correct.

## What changed

- Track the source page associated with each extracted XObject.
- Read a direct `/DeviceRGB` or `/DeviceCMYK` value from the source page's transparency group.
- Register and remap only the RGB and CMYK group color spaces that are actually needed.
- Omit the outer group `/CS` when the source page does not provide a supported explicit value.
- Add regression fixtures and tests for:
  - explicit RGB page groups,
  - explicit CMYK page groups,
  - CMYK content without an explicit page-group color space,
  - spot-color content using `/Separation /SpotGreen` without an explicit page-group color space.

## Scope

This is intentionally not a full color-management or PDF/X implementation. It does not add ICC profile selection, convert document colors, or infer a single color space for mixed-content PDFs. Broader work is tracked in [typst/typst#3143](https://github.com/typst/typst/issues/3143) and [typst/typst#6012](https://github.com/typst/typst/issues/6012).

## Tests

```bash
cargo test -p tests pdf_embedded_as_xobject_ -- --nocapture
cargo test -p krilla --features pdf
cargo test -p tests
cargo fmt --check
git diff --check
```


LaurenzV:
Hmm, thanks, but just using a heuristic to determine whether a PDF is "RGB or CMYK" is not a good idea I think. The whole colorspace thing was introduced this release due to a different but report in Typst, and it seemed like Latex does the same. But I tried it with your PDFs and it doesn't always seem to be the case, so I need to look into how exactly Latex works in this regard.

Hyperrick:
> Hmm, thanks, but just using a heuristic to determine whether a PDF is "RGB or CMYK" is not a good idea I think. The whole colorspace thing was introduced this release due to a different but report in Typst, and it seemed like Latex does the same. But I tried it with your PDFs and it doesn't always seem to be the case, so I need to look into how exactly Latex works in this regard.

Hey Laurenz,

thanks for the reply!

I tested this again with actual print-production PDFs locally. In Acrobat Output Preview the previous output I was testing still broke the separations: black was no longer preserved correctly after export. The problematic case was that the outer embedded PDF XObject got an inferred `/CS`, instead of leaving it unspecified.

I also checked what pdflatex/lualatex do with the same kind of input. They don’t seem to infer CMYK from the page contents. They leave the outer group without a `/CS`, while the nested CMYK/Separation groups from the embedded PDF stay intact.

I changed the patch in that direction now:
- keep an explicit source page `/Group /CS` if there is one,
- otherwise don’t write `/CS` on the outer group,
- don’t try to guess from CMYK/Separation/DeviceN usage.

This fixed the Acrobat Output Preview issue in my local production PDFs, and should line up better with the behavior of pdflatex/lualatex.

Hyperrick:
Hi @LaurenzV — gentle follow-up on this. I reworked the patch to address your concern: it no longer infers RGB or CMYK from page contents. It now preserves an explicit source `/Group /CS` for `/DeviceRGB` and `/DeviceCMYK`, and otherwise leaves `/CS` unset.

I also updated the PR description to reflect the revised implementation and its relationship to the existing regression case from typst/typst#7269. The targeted XObject tests pass locally. The GitHub Actions run is currently waiting for maintainer approval.

Would you mind taking another look and approving the workflow when you have a chance? Thanks!

LaurenzV:
Sorry about that, haven't forgotten about it but I procrastinated looking into it again. 😅 After looking into it again, I think you are right. What pdflatex seems to do is just preserve the group object of the original page. I hadn't seen that the original PDF from the typst issue had such a group. So we should just do the same.

However, krilla is likely wrong place for this fix. I will fix this upstream in `hayro-write`.

Hyperrick:
Thanks for taking another look! That makes sense — preserving the original group object in `hayro-write` sounds like the right layer for this. I’d be happy to test the upstream fix against my print-production PDFs once it’s available.

LaurenzV:
Can you test again now?

Hyperrick:
Retested now with both the synthetic CMYK/spot/RGB fixtures and our own production PDFs. I can confirm the bug is fixed:

- the original embedded page streams are preserved byte-for-byte,
- no synthetic RGB group is added when the source page has no group,
- the CMYK measurements match the source again, including K-only black staying `C0 M0 Y0 K100`,
- the `SpotGreen` separation remains intact,
- and the RGB control case is unchanged.

Thanks a lot for taking another look, implementing the proper fix in `hayro-write`, and updating the branch — much appreciated!

LaurenzV:
Thanks!

PR #419, opened 2026-07-31
Support custom document information fields

danbishop-canva:
PDF document information dictionaries can contain application-defined entries, but `Metadata` currently exposes only the standard fields.

Add `Metadata::custom_field` for writing custom text entries through `pdf-writer`'s document information dictionary API. Repeated field names replace their earlier values, while empty names and names reserved for standard document information fields are ignored.

Custom information fields remain separate from XMP metadata. Custom XMP properties and namespace handling are covered independently by #408.

### DIFF
--- a/crates/krilla/src/interchange/metadata.rs
+++ b/crates/krilla/src/interchange/metadata.rs
@@ -130,7 +130,7 @@ impl Metadata {
     /// document information fields.
     ///
     /// **WARNING**: Custom fields will currently be ignored when exporting to
-    /// PDF/A-2 and PDF/A-3. This is a temporary limitation that will be lifted
+    /// PDF/A-2 and PDF/A-3 and PDF/A-4. This is a temporary limitation that will be lifted
     /// in the future.
     #[must_use]
     pub fn custom_field(mut self, name: String, value: String) -> Option<Self> {
--- a/crates/krilla/src/interchange/metadata.rs
+++ b/crates/krilla/src/interchange/metadata.rs
@@ -126,16 +126,19 @@ impl Metadata {
     /// Add a custom field to the document information dictionary.
     ///
     /// If a field with the same name was already added, its value will be
-    /// replaced. Empty names and names reserved for standard document
-    /// information fields are ignored. Custom fields are not included in the
-    /// XMP metadata. PDF/A-2 and PDF/A-3 readers may ignore the document
-    /// information dictionary, and export modes that prohibit the dictionary
-    /// will omit these fields.
-    pub fn custom_field(mut self, name: String, value: String) -> Self {
-        if !name.is_empty() && !is_standard_document_info_field(&name) {
-            self.custom_fields.insert(name, value);
+    /// replaced. Returns `None` for empty names and names reserved for standard
+    /// document information fields. 
+    /// 
+    /// Currently, custom fields are not included in the XMP
+    /// metadata and will fail export in  PDF/A-2 and PDF/A-3.
+    #[must_use]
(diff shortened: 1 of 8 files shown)


LaurenzV (commented):
Overall LGTM, but have you checked whether PDF-A/1-3 permit having additional metadata in the document dictionary that is not present in the XML metadata? At least from a cursory read it seems like the two must be consistent.

If we are planning to fix this as part of #408, it's probably fine to just add a TODO somewhere for now. Let me know if I'm missing something.

danbishop-canva:
@LaurenzV from my understanding PDF/A-1 permits additional metadata in the document dictionary that is not present in the XML metadata, but PDF/A-2/3 do not. 

Yes I think this will be something that will be fixed by #408. I think its best to get #408 merged and then I can apply some work around validation on this PR.

LaurenzV:
Could you add the same custom test but with PDF/A-1 and PDF/A-3? Let's see whether veraPDF complains in CI. If so, we can just add an ignore to the test for now and then unignore it when the other PR is merged. If you need it, it would be fine to land this PR before the other one, up to you.

reknih:
**PDF/A-1:**

> If [the document information dictionary] does appear, then all of its entries that have analogous properties in predefined XMP schemas, as defined by Table 1, shall also be embedded in the file in XMP form with equivalent values. Any document information dictionary entry not listed in Table 1 shall not be embedded using a predefined XMP schema property.

Put anything in there and you're good

**PDF/A-2 and 3:**

> A document information dictionary may be present in a conforming file and a PDF/A-2 compliant reader shall ignore it.

> A document information dictionary may appear within a conforming file. If it does appear, a compliant PDF/A-2 reader shall ignore it. A PDF/A-2 conforming writer should ensure that the values in the document information dictionary are consistent with the corresponding values in the document’s metadata stream as listed in Table 7.

So that seems to me like as long as the metadata is reflected in both, you are doing good. That would mean raising a `ValidationError` if no custom XMP is present

**PDF/A-4:**

> If a document information dictionary is present, it shall only contain a ModDate entry.

Means that in PDF/A-4, it's forbidden. It's also deprecated in PDF 2.0:

> As described in 14.3.3, "Document information dictionary", this method for specifying document metadata has been deprecated in PDF 2.0 and should therefore only be used to encode information that is stated as required elsewhere in this document.

> Earlier versions of the PDF file format used the document information dictionary to represent document level metadata. In PDF 2.0 such use is deprecated except for two entries, `CreationDate` and `ModDate`. For any other document level metadata, a metadata stream (see 14.3.2 "Metadata streams") should be used instead.

crates/krilla/src/interchange/metadata.rs
```diff
@@ -122,6 +123,21 @@ impl Metadata {
...
+    /// Add a custom field to the document information dictionary.
+    ///
+    /// If a field with the same name was already added, its value will be
+    /// replaced. Empty names and names reserved for standard document
+    /// information fields are ignored. Custom fields are not included in the
+    /// XMP metadata. PDF/A-2 and PDF/A-3 readers may ignore the document
+    /// information dictionary, and export modes that prohibit the dictionary
+    /// will omit these fields.
+    pub fn custom_field(mut self, name: String, value: String) -> Self {
+        if !name.is_empty() && !is_standard_document_info_field(&name) {
```
reknih: I think that this could return `None` or an error variant of a `Result` if the value has been ignored.

LaurenzV:
Thanks! I've decided to just ignore them for PDF/A-2 and PDF/A-3 for now, and added a bit warning in the doc comment.

PR #423, opened 2026-08-11
Add support for forms (part 1: push buttons, checkboxes and radio groups)

diogotcorreia:
~~Very wip still, but opening to gather early feedback.~~ **EDIT 2026-08-25:** should be ready now

Notably, I'm looking for feedback on:
- how I'm linking the field with the widget annotations (EDIT: this has been changed in commit https://github.com/LaurenzV/krilla/pull/423/commits/c7890ee9eff96bce10d1927095cbd3bd2df8ab6e)
- ~~how I'm abstracting the fields into a flat structure (or if I should be doing that at all)~~ (EDIT: this has been changed in commit https://github.com/LaurenzV/krilla/pull/423/commits/ee7abb36da1015fda64e7d919d281adf34f7c977 to use a tree structure now)
- overall public interface (see example below)

~~I'm currently testing with the following code, which also gives an overview of the proposed API: https://gist.github.com/diogotcorreia/2d4cf96def0dbf3817b15884cda638ba~~ See example in `crates/krilla/examples/forms.rs` instead.
As of the time of writing, compiling this PR requires pdf-writer with PR https://github.com/typst/pdf-writer/pull/86 applied.

TODO (no need to comment on these yet):
- [x] doc comments
- [x] tests
- [x] PDF-A/UA support
- [x] PDF version gating
- [x] example (thanks @reknih for the suggestion)

---

Relates to https://github.com/typst/typst/issues/2421


reknih:
It would be nice if alongside your implementation, you could provide an example that grows along with it to serve as a reviewer entrypoint for understanding the API and its intended usage!

crates/krilla/src/interactive/forms.rs
```diff
@@ -0,0 +1,420 @@
...
+    annotation::{DualStateAppearance, SimpleAppearance, WidgetAnnotation},
+    chunk_container::ChunkContainer,
+    geom::Rect,
+    serialize::SerializeContext,
+    stream::Stream,
+    tagging::AnnotationIdentifier,
+};
+
+#[derive(Default)]
+pub(crate) struct AcroForm {
```
saecki: Can we get rid of this wrapper struct? It doesn't seem to really add anything.

crates/krilla/src/serialize.rs
```diff
@@ -1081,6 +1093,8 @@ pub(crate) struct GlobalObjects {
     outline: MaybeTaken<Option<Outline>>,
     /// Stores the tag tree.
     tag_tree: MaybeTaken<Option<TagTree>>,
+    /// Stores the form fields.
+    pub(crate) forms: MaybeTaken<AcroForm>,
```
saecki: I think for consistency and to avoid serializing an empty interactive forms dictionary, when no `FieldTree` is added, this should be an `Option`.

crates/krilla/src/interactive/form.rs
```diff
@@ -0,0 +1,420 @@
...
+    Leaf(FieldKind),
+}
+
+impl Node {
+    fn serialize_node(
+        &self,
+        sc: &mut SerializeContext,
+        chunk_container: &mut ChunkContainer,
+    ) -> Ref {
+        match self {
```
saecki, line 180: All fields that aren't immediate children of the root form should have the `Parent` key set:

> (Required if this field is the child of another in the field hierarchy;
absent otherwise) The field that is the immediate parent of this one
(the field, if any, whose Kids array includes this field). A field can have
at most one parent; that is, it can be included in the Kids array of at
most one other field.

crates/krilla/src/page.rs
```diff
@@ -238,6 +240,43 @@ impl<'a> Page<'a> {
         }
     }
 
+    /// Add a widget annotation to the page.
+    ///
+    /// The given field must be the same where annotation was created from.
```
saecki, line 245: The `new_widget` functions seem to take `&self` for no reason, other than ensuring the widget annotation has the right type. But since that type information isn't relevant when adding it later on with this method, I think this just adds unnecessary friction.

If we decide to go with the current API of adding the widget annotation to the page, `new_widget` functions should be moved to `WidgetAnnotation` IMO.

It would be nice to have a type-safe builder function, but I currently don't see a way where we can use it when adding the widget annotation to the form field.

crates/krilla/src/interactive/annotation.rs
```diff
@@ -275,9 +280,327 @@ impl LinkAnnotation {
...
+pub(crate) trait SerializeAppearance {
+    // TODO: refactor, this is the only way I could find to make the borrow checker happy
+    type RefsHolder;
+
+    fn register_refs(
+        self,
+        sc: &mut SerializeContext,
+        chunk_container: &mut ChunkContainer,
+    ) -> Self::RefsHolder;
+    fn serialize_appearance(
```
saecki, line 642: Could you add a blank line here?

crates/krilla/src/page.rs
```diff
@@ -238,6 +240,43 @@ impl<'a> Page<'a> {
...
+    /// Passing a different field is a logic error and may result in an invalid PDF.
+    pub fn add_widget_annotation<FT, AT>(
```
saecki: One limitation of taking a `WidgetAnnotation` instead of an `Annotation` here, is that there is now way to add a `Location` (a Typst `Span`), which can be used in an error message. Though I see why the more explicit type is desirable 🤔
Instead we could just assert that the added `Annotation` is a widget annotation.

crates/krilla/src/interactive/forms.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// Takes its position on the page, as well as its appearance.
+    pub fn new_widget(&self, rect: Rect, appearance: Stream) -> WidgetAnnotation<SimpleAppearance> {
+        WidgetAnnotation::simple(rect, appearance)
+    }
+}
+
+impl FormField<kind::Checkbox> {
+    /// Create a checkbox field.
+    /// The field name must not contain any dot character (`.`).
+    pub fn checkbox(name: String) -> Self {
```
saecki: Maybe we should add a check to form field constructors to validate the `name` doesn't contain a `.` character.

crates/krilla/src/page.rs
```diff
@@ -238,6 +240,43 @@ impl<'a> Page<'a> {
...
+        field: &mut FormField<FT>,
+        mut annotation: WidgetAnnotation<AT>,
+    ) -> Identifier
+    where
+        WidgetAnnotation<AT>: Into<Annotation>,
+    {
+        let parent_ref = match field.identifier {
+            Some(ref_) => ref_,
+            None => {
+                let ref_ = self.sc.new_ref();
```
saecki, line 257: This is a bit nitpicky, but I think currently only page refs are created before final serialization of the document, which meant their refs were contiguous. By creating the form field refs ad-hoc, they are interspersed with the page refs.

crates/krilla/src/page.rs
```diff
@@ -238,6 +240,43 @@ impl<'a> Page<'a> {
         }
     }
 
+    /// Add a widget annotation to the page.
+    ///
+    /// The given field must be the same where annotation was created from.
+    /// Passing a different field is a logic error and may result in an invalid PDF.
+    pub fn add_widget_annotation<FT, AT>(
+        &mut self,
+        field: &mut FormField<FT>,
```
saecki, line 251: I'm still not sure if requiring the `FormField` here might be annoying when implementing things from the Typst side. Maybe we can just try it out this way, and come back to this later when it becomes an issue.

crates/krilla/src/interactive/form.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+#[allow(private_bounds)]
+impl<T: SerializableField> FormField<T> {
+    fn serialize_field(
+        &self,
+        sc: &mut SerializeContext,
+        chunk_container: &mut ChunkContainer,
+    ) -> Ref {
+        let root_ref = self.identifier.unwrap_or_else(|| sc.new_ref());
```
saecki, line 465: By moving the general code into the `FieldKind::serialize_field` method above the match statement, I think we can eliminate the `SerializableField` trait below.

crates/krilla/src/interactive/forms.rs
```diff
@@ -0,0 +1,442 @@
...
+            });
+            field.children(annotations);
+        }
+
+        root_ref
+    }
+}
+
+/// A type-agnostic field.
+#[derive(Debug, Clone)]
```
saecki: I think moving this enum above the `FormField` struct makes more sense.

crates/krilla/src/interactive/forms.rs
```diff
@@ -0,0 +1,442 @@
...
+
+            field
+                .field_type(pdf_writer::types::FieldType::Button)
+                .radio_value(Name(value))
+                .radio_default_value(Name(default_value));
+        }
+    }
+}
+
+impl From<FormField<kind::PushButton>> for FieldKind {
```
saecki: Can you move the `From` impls below the inherent impls of each type.

saecki (commented):
This is already looking quite promising :)
I've left a few suggestions and thoughts.

crates/krilla/src/interactive/annotation.rs
```diff
@@ -275,9 +280,352 @@ impl LinkAnnotation {
...
+            }
+            WidgetAnnotationKind::DualState(a) => {
+                a.serialize_type(sc, chunk_container, root_ref, page_height)
+            }
+        }
+    }
+}
+
+trait SerializableAppearance {
+    // TODO: refactor, this is the only way I could find to make the borrow checker happy
```
diogotcorreia: Not sure if the current approach is too hacky or not. The problem here is that we need a mut reference to chunk_container (for `sc.register_cacheable`) while we are writing to the annotations chunk.
So we need to register the appearance refs before we start writing the annotation

crates/krilla/src/interactive/form.rs
diogotcorreia: Seems like I accidentally removed that code when refactoring to a tree structure in https://github.com/LaurenzV/krilla/commit/ee7abb36da1015fda64e7d919d281adf34f7c977
Will add it back, thanks for noticing.

crates/krilla/src/interactive/form.rs
diogotcorreia: I think it makes the code a lot more convoluted since we can't access the struct fields in `FieldKind` (e.g., `self.name`, `self.alt_name`, etc.). I could make a bunch of getters, but I think that's worse

diogotcorreia: I can get rid of the trait if I do something like this instead:

```diff
diff --git a/crates/krilla/src/interactive/forms.rs b/crates/krilla/src/interactive/forms.rs
index 3410397..94c22aa 100644
--- a/crates/krilla/src/interactive/forms.rs
+++ b/crates/krilla/src/interactive/forms.rs
@@ -5,7 +5,11 @@
 //! multiple fields. These fields are organized in a tree structure.
 //! Each field can be displayed on the page via one or more annotations.
 
-use pdf_writer::{types::FieldFlags, writers::Form, Finish, Ref, TextStr};
+use pdf_writer::{
+    types::FieldFlags,
+    writers::{Field, Form},
+    Finish, Ref, TextStr,
+};
 
 use crate::{
     annotation::{DualStateAppearance, SimpleAppearance, WidgetAnnotation},
@@ -286,13 +290,13 @@ impl FormField<kind::Radio> {
 }
 
 #[allow(private_bounds)]
-impl<T: SerializableField> FormField<T> {
-    fn serialize_field(
+impl<T> FormField<T> {
+    fn serialize_field<'a>(
         &self,
         sc: &mut SerializeContext,
-        chunk_container: &mut ChunkContainer,
+        chunk_container: &'a mut ChunkContainer,
         parent_ref: Option<Ref>,
-    ) -> Ref {
+    ) -> (Ref, Field<'a>) {
         let root_ref = self.identifier.unwrap_or_else(|| sc.new_ref());
         let mut field = chunk_container.non_stream.fields.form_field(root_ref);
 
@@ -312,8 +316,6 @@ impl<T: SerializableField> FormField<T> {
             field.mapping_name(TextStr(mapping_name));
         }
 
-        self.kind.serialize_field(&mut field);
-
         if !self.annotations.is_empty() {
             let annotations = self.annotations.iter().map(|identifier| {
                 let page_annotations = sc.page_infos()[identifier.page_index].annotations();
@@ -322,7 +324,7 @@ impl<T: SerializableField> FormField<T> {
             field.children(annotations);
         }
 
-        root_ref
+        (root_ref, field)
     }
 }
 
@@ -345,30 +347,36 @@ impl FieldKind {
         parent_ref: Option<Ref>,
     ) -> Ref {
         match self {
-            Self::PushButton(f) => f.serialize_field(sc, chunk_container, parent_ref),
-            Self::Checkbox(f) => f.serialize_field(sc, chunk_container, parent_ref),
-            Self::Radio(f) => f.serialize_field(sc, chunk_container, parent_ref),
+            Self::PushButton(f) => {
+                let (ref_, mut field) = f.serialize_field(sc, chunk_container, parent_ref);
+                f.kind.serialize_field(&mut field);
+                ref_
+            }
+            Self::Checkbox(f) => {
+                let (ref_, mut field) = f.serialize_field(sc, chunk_container, parent_ref);
+                f.kind.serialize_field(&mut field);
+                ref_
+            }
+            Self::Radio(f) => {
+                let (ref_, mut field) = f.serialize_field(sc, chunk_container, parent_ref);
+                f.kind.serialize_field(&mut field);
+                ref_
+            }
         }
     }
 }
 
-pub(crate) trait SerializableField {
-    fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>);
-}
-
 /// Field kind structs.
 pub mod kind {
     use pdf_writer::{types::CheckBoxState, Name};
 
-    use super::SerializableField;
-
     /// A push button.
     /// Create a field of this type via [`FormField::push_button`](super::FormField::push_button).
     #[derive(Debug, Clone, Default)]
     pub struct PushButton;
 
-    impl SerializableField for PushButton {
-        fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
+    impl PushButton {
+        pub(super) fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
             field.field_type(pdf_writer::types::FieldType::Button);
         }
     }
@@ -381,8 +389,8 @@ pub mod kind {
         pub(super) default_checked: Option<bool>,
     }
 
-    impl SerializableField for Checkbox {
-        fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
+    impl Checkbox {
+        pub(super) fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
             let value = if self.checked.unwrap_or(false) {
                 CheckBoxState::Yes
             } else {
@@ -409,8 +417,8 @@ pub mod kind {
         pub(super) default_value: Option<String>,
     }
 
-    impl SerializableField for Radio {
-        fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
+    impl Radio {
+        pub(super) fn serialize_field<'a>(&self, field: &mut pdf_writer::writers::Field<'a>) {
             let value = if let Some(value) = &self.value {
                 value.as_bytes()
             } else {
```

The match statement gets a bit worse, but can probably be alleviated by using a macro. Or I can pass a `FnOnce` that calls `self.kind.serialize_field`. What do you prefer?

I'm quite inclined to do the `FnOnce` since iirc I've seen that pattern in krilla quite a lot.

crates/krilla/src/page.rs
diogotcorreia: That's probably true, but there is a remapping step in chunk container that re-numbers everything anyway, so the resulting PDFs do not have their page refs interspersed with annotations.
https://github.com/LaurenzV/krilla/blob/fb07499278a2d2524dda38c41f338dec28bd0e48/crates/krilla/src/chunk_container.rs#L107-L111

crates/krilla/src/page.rs
diogotcorreia: We still have one problem with the current API: in certain cases, the widget annotation needs to know the value of the field.
https://github.com/diogotcorreia/krilla/blob/65dc460ec476fe5015b43c608677e165d38e8d98/crates/krilla/src/interactive/annotation.rs#L601-L602

This is true for checkboxes/radio groups, but I suspect it might be the case for variable text fields as well.

I was thinking of solving this by forcing the value to be passed in the constructor of each field (to ensure it is immutable) and pass it on to the annotation via `new_widget`. However, moving `new_widget` to `WidgetAnnotation` would break this.

I'm wondering then how we should handle this? Should be make the API more brittle by requiring the caller to keep the widget and field values in-sync?

crates/krilla/src/page.rs
diogotcorreia: I haven't looked into Typst yet, but I can see why passing `Location` is important. I made the function take an `Annotation` instead and panic if it's not a widget.

crates/krilla/src/interactive/forms.rs
```diff
@@ -0,0 +1,442 @@
...
+            name,
+            flags: FieldFlags::RADIO,
+            ..Default::default()
+        }
+    }
+
+    /// Set the value of the radio group.
+    /// It should correspond to a value of one of the annotations.
+    // TODO: take Option<String> instead?
+    pub fn set_value(&mut self, value: String) {
```
diogotcorreia: I guess `Option<String>` would be better since it allows to set the back group to unchecked, even though it's the default?

diogotcorreia (commented):
Thank you @saecki!
I've addressed most of your comments and left a reply on the others

crates/krilla/src/interactive/annotation.rs
```diff
@@ -275,9 +280,360 @@ impl LinkAnnotation {
...
+        annotation: &mut pdf_writer::writers::Annotation<'_>,
+        refs: Self::RefsHolder,
+    );
+
+    fn register_appearance_entry(
+        sc: &mut SerializeContext,
+        chunk_container: &mut ChunkContainer,
+        appearance: Stream,
+    ) -> Ref {
+        let xobject = XObject::new(appearance, false, false, None);
```
diogotcorreia: While writing an example I noticed the appearance stream has its y-axis flipped compared to the rest of the document (i.e., text is upside down). It seems to be a krilla convention. Should we also flip it here?

crates/krilla/src/page.rs
saecki: If the current API helps solving a problem then I think it's fine. In the current state I just didn't see any reason for it to be this way :)

crates/krilla/src/page.rs
saecki: Then forget what I've said.

crates/krilla/src/interactive/form.rs
saecki: Hmm, that doesn't really seem nicer. Let's just keep it as it then.

crates/krilla/src/interactive/annotation.rs
saecki: I think the two-step approach is fine. The nested four tuple just seems a little messy. Maybe you can add a struct for that one. Also the `register_appearance_entry()` trait method can just be a top-level function.

crates/krilla/src/interactive/annotation.rs
saecki: Yes, I think it should be flipped inside the widget annotations rectangle. Maybe it could be done similar to how it's done for patterns.

crates/krilla/src/interactive/forms.rs
saecki: Yes, that seems fair.

Also regarding the setters, krilla often provides builder methods prefixed with `with_` that take `self` by value, which would be convenient when building widget annotations and forms.

crates/krilla/src/interactive/form.rs
```diff
@@ -0,0 +1,533 @@
...
+        chunk_container: &mut ChunkContainer,
+        parent_ref: Option<Ref>,
+    ) -> Ref {
+        let root_ref = self.identifier.unwrap_or_else(|| sc.new_ref());
+        let mut field = chunk_container.non_stream.fields.form_field(root_ref);
+
+        let mut flags = self.flags;
+        if sc.serialize_settings().pdf_version() < PdfVersion::Pdf15 {
+            flags.remove(FieldFlags::RADIOS_IN_UNISON);
+        }
```
diogotcorreia, line 471: Should this be implemented with a `VersionedFeature` instead? And throw an error if not available?

crates/krilla/src/configure/validate.rs
```diff
@@ -123,6 +123,8 @@ pub enum ValidationError {
     MissingDocumentOutline,
     /// An annotation is missing an alt text.
     MissingAnnotationAltText(Option<Location>),
+    /// A form field is missing an alt name.
+    MissingFieldAltName(Option<Location>),
```
reknih, line 127: How does this error differ from `MissingAnnotationAltText`?

crates/krilla/src/configure/validate.rs
```diff
@@ -548,6 +557,9 @@ impl Archival {
                 | ValidationError::Transparency(_)
                 | ValidationError::ImageInterpolation(_)
                 | ValidationError::EmbeddedFile(EmbedError::Existence, _)
+                | ValidationError::AnnotationHasConditionalAppearance(_)
```
reknih, line 565: PDF/A-1 has a stricter clause than the other parts: Push buttons, like other annotations, cannot contain appearance subdictionaries for their values. I suggest modelling this using an `is_button` field in the enum variant.

crates/krilla/src/configure/validate.rs
```diff
@@ -135,6 +137,13 @@ pub enum ValidationError {
     EmbeddedFile(EmbedError, Option<Location>),
     /// The PDF contains no tagging.
     MissingTagging,
+    /// The PDF contains an annotation with a rollover or down appearance.
+    AnnotationHasConditionalAppearance(Option<Location>),
+    /// The PDF contains an action that can mutate the appearance of the document (e.g., multimedia
+    /// content, reset form, etc.).
+    ContainsMutatingAction(Option<Location>),
+    /// The PDF contains an annotation or field that has an additional-actions dictionary.
+    ContainsAdditionalActions(Option<Location>),
```
reknih, line 148: In PDF/A-4, Section 6.6.3, the trigger events for the document catalog and pages are restricted with a _should not_ and then, below, there is a hard requirement for the keys allowed should the dictionary be present anyways. I assume you only apply this validation error to form fields right now? Without a comment, it is easy to make errors here in the future, should we add `AA` on document catalogues or page dictionaries.

crates/krilla/src/interactive/action.rs
```diff
@@ -66,3 +78,41 @@ impl LinkAction {
             .uri(Str(self.uri.as_bytes()));
     }
 }
+
+/// A reset form action. Will reset the given form fields when triggered.
+pub enum ResetFormAction {
```
reknih, line 86: Are derives missing here?

crates/krilla/src/interactive/annotation.rs
```diff
@@ -44,6 +46,22 @@ impl Annotation {
         }
     }
 
+    /// Create a new widget annotation with some alt text.
+    ///
+    /// Note that the alt text might be required in some cases, for example
+    /// when exporting to PDF/UA.
+    pub fn new_widget(
+        annotation: impl Into<WidgetAnnotationKind>,
```
reknih:
```suggestion
        widget: impl Into<WidgetAnnotationKind>,
```
?

crates/krilla/src/interactive/annotation.rs
```diff
@@ -62,37 +80,32 @@ impl From<LinkAnnotation> for Annotation {
     }
 }
 
+impl From<WidgetAnnotationKind> for Annotation {
+    fn from(value: WidgetAnnotationKind) -> Self {
```
reknih:
```suggestion
    fn from(widget: WidgetAnnotationKind) -> Self {
```

crates/krilla/src/interactive/annotation.rs
```diff
@@ -275,9 +304,555 @@ impl LinkAnnotation {
...
+
+    /// Set the appearance of the annotation when the mouse
+    /// is pressed or is being held down over the annotation's area.
+    pub fn with_down_appearance(mut self, appearance: Stream) -> Self {
+        self.set_down_appearance(appearance);
+        self
+    }
+}
+
+/// Applies to widget annotations with a single state appearance.
```
reknih: Dupe comment

crates/krilla/src/interactive/annotation.rs
reknih: Where is this implementation used? Setting the parents and location to None seems a bit dangerous

reknih (commented):
@LaurenzV would you like a `pdf-writer` release before merging this or are you also fine setting it to a Git dependency for now?

LaurenzV:
Git sounds fine for now I think!

diogotcorreia:
> Git sounds fine for now I think!

Note that https://github.com/typst/pdf-writer/commit/d2a29d5fc74001650aaa329e1507184b48f8b2ea will cause test failures, so that should probably be handled in a separate PR. I can do it in a sec.

LaurenzV:
Separate would be preferable yes!

crates/krilla/src/configure/validate.rs
diogotcorreia: This refers to the `TU` attribute (PDF 2.0 Table 226) vs the `Contents` attribute (PDF 2.0 Table 166) for `MissingAnnotationAltText`. It is slightly confusing as I think only one of them is required for PDF/UA-1, but we are currently requiring both of them to be set.

crates/krilla/src/configure/validate.rs
diogotcorreia: Currently I made `AA` allowed under PDF/A-4 (and forbidden in all other PDF/A versions). There is nothing related to the allowed keys, since we only use `D` in forms anyway, which is allowed in PDF/A-4.

I'll leave a comment noting this then.

crates/krilla/src/configure/validate.rs
diogotcorreia: I believe PDF/A-1 Corr 2 (page 4) changes that paragraph to match PDF/A-2 and later.

crates/krilla/src/interactive/action.rs
diogotcorreia: `Action`, `LinkAction` and `Destination` also don't have them (except for `Destination` that has `Hash`), so I didn't add it.

But I don't see why not, I can add it to all of them.

crates/krilla/src/interactive/annotation.rs
diogotcorreia: Sure. It was copied from the `new_link` function above, but I agree it's better.

crates/krilla/src/interactive/annotation.rs
diogotcorreia: I believe it was used in `page.add_widget_annotation` before, but now that just takes an `Annotation`. Right now it is only used in tests.

This impl also exists for `LinkAnnotation` tho, so that is why I also added it. It is basically a shorthand for `Annotation::new_widget(..., None)`.

I don't think it's that dangerous, as `struct_parent` is set when `page.add_widget_annotation ` is called, and having a missing `location` doesn't seem problematic since it's just for error reporting.

crates/krilla/src/interactive/annotation.rs
diogotcorreia: Nice catch, fixed

crates/krilla/src/configure/validate.rs
reknih: Can you check screen reader (for pairings, see accessibility guide) and validator (veraPDF) behavior? I would like to prevent the same alternate description being read twice.

crates/krilla/src/interactive/annotation.rs
reknih: Got it

crates/krilla/src/configure/validate.rs
diogotcorreia: veraPDF does not care about the `Contents` attribute for widget annotations, only `TU` or the `Alt` property of the structure element: https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFUA-Part-1-rules#rule-7181-3

I tried Adobe Reader with NVDA and got the following (haven't used a screen reader before tho, so not sure if I did anything wrong):

- The `Contents` attribute is completely ignored
- If widget annotation has both `TU` and `Alt`, then `Alt` is read when the page is being read, but `TU` is used when tabbing through the forms
- If only `TU` is present, then it is used when reading the page as well
- If only `Alt` is present, it is used when the page is being read, but not used when tabbing (it only says e.g., "checkbox, not checked").


So I'm going to drop `Contents` from widget annotations, as it doesn't appear to be used at all.
I'm not sure if checking for `Alt` in the structure elements only if `TU` is missing (and vice-versa) is straightforward in krilla, but it doesn't appear to be. I would suggest that we mandate `TU` to be present (in light of the results above), but if that's unacceptable I can see if there is a way to check for `TU` OR `Alt`.

---

Used code for future reference: https://bin.diogotc.com/K3+qnwzfsqb.rs

diogotcorreia: I removed the `Annotation::new_widget` function altogether since it would have been equivalent to `Annotation::from`, but I can restore it if you think it makes sense to have an explicit function for it.

diogotcorreia:
Tests failing seem to be related to the buttons not containing the `DA` property. This is only defined in the spec under "variable text", which I would argue buttons are not, but there appears to be some discussion upstream: https://github.com/veraPDF/veraPDF-arlington-tools/issues/13#issuecomment-3106347509

What should we do? :thinking:

crates/krilla/src/configure/validate.rs
reknih: IMO requiring `TU` is fine.

reknih:
Given https://github.com/pdf-association/pdf-issues/issues/323#issuecomment-1954348054, the treatment of `DA` for radios etc. is underspecified. IMO, ISO 32000-2:2020, Clause 12.7.4.3, the note makes the intent clear to be that only fields containing user-defined text are Variable Text fields:

> Examples include text fields to be filled in with text typed by the user from the keyboard, scrollable list boxes whose contents are determined interactively at the time the document is displayed and fields containing current dates or values calculated by an ECMAScript.

Thus, radio buttons, checkboxes, and buttons should not contain keys from Table 228, including `DA`. I also would not know what the dynamic appearance of a checkbox or button should be in practice. So if it does not mess with compatibility, I would omit the key as you do now.

With regards to the tests, can you check if the newest available Arlington checker also reports this issue? If so, I suggest to disable that specific rule for these specific tests. If that's not possible, exclude the files from Arlington checking altogether.

crates/krilla/src/interactive/form.rs
reknih: This flag causes a behavioral change for the forms user, so I think the form's intent is not conserved if it cannot be set. So I think erroring here is better (that's a Result, not a panic, right?)

diogotcorreia:
> With regards to the tests, can you check if the newest available Arlington checker also reports this issue? If so, I suggest to disable that specific rule for these specific tests. If that's not possible, exclude the files from Arlington checking altogether.

I could not find a way to disable specific rules, and I checked that it is still present in version 1.31.174.
Therefore, I just excluded the files from the Arlington check.

reknih (approved):
LGTM from my POV

reknih:
@LaurenzV from our side, this would be ready

LaurenzV (commented):
Cool, thanks so much for the contribution! I don't think I have the time to really review this, all I can offer is some review comments from AI. 😅 Let me know if these are irrelevant or not, sorry if it's just slop or inaccurate comments! Otherwise, happy to merge.

image

diogotcorreia:
@LaurenzV

1. This is kinda true, the `ValidationError::RequiresNewerPdfVersion` only runs when the UA1 validator is set (which I had on during my testing, so I didn't notice). Do you know if there is another mechanism with the same semantics but that does not care about the `UA`/`A` validators? I couldn't find anything sadly, and a panic seems inappropriate, I'd prefer if we could catch the error in typst and display an appropriate message.
   The last two sentences are hallucinated tho, as you cannot set the RADIOS_IN_UNISON flag for non-radio groups.
2. Fixed; didn't notice because firefox/pdfium don't seem to support rolldown/down appearances anyway, but tested with Acrobat now
3. This is true, but it does sound more like a bug in pdfium than something we should fix. Given our current API design, this does not seem that easy to implement unless we just make a `NamedAppearanceStream` behave like a `SimpleAppearanceStream` when not exporting to PDF/A. Thoughts? (PS: I cannot click the link to the source code through the image :laughing: could you post it?) **EDIT:** opened pdfium bug https://issues.chromium.org/issues/561372468
4. This might need a bit of a refactor, I'll check if it's possible with the current API design
5. Fair, I never thought this as a supported use-case, so I'll remove the Clone and hopefully that doesn't cause any issues.
6. Seems like something that should be checked by typst too, but I'll add an `assert_debug!` and a mention on the docs. I need to check if adding support for `Opt` would remove this restriction tho, it seems like it might.

diogotcorreia:
Seems like the experimental `git history split` command doesn't sign commits yet :sweat_smile: TIL

LaurenzV:
> Do you know if there is another mechanism with the same semantics but that does not care about the UA/A validators? I couldn't find anything sadly

Feel free to just add a TODO then!

diogotcorreia:
I believe this should be ready now :crossed_fingers:

LaurenzV:
Thanks!!

PR #427, opened 2026-08-21
Fix typos in comments

diogotcorreia:
Keeping as draft in case I find more while I am working on pdf forms


LaurenzV:
Is this ready or are do you think there will be more typos? 😄

PR #430, opened 2026-09-15
Add support for forms (part 2: text and choice fields)

diogotcorreia:
Continuation of #423

## TODO

- [x] Rebase and cleanup commit history

---

Relates to https://github.com/typst/typst/issues/2421


diogotcorreia:
CI failure seems to be a bug in pdfcpu: https://github.com/pdfcpu/pdfcpu/issues/1488

crates/krilla/examples/forms.rs
```diff
@@ -37,6 +40,48 @@ fn main() {
...
+    // Set an alternative name, for accessibility purposes.
+    combobox_field.set_alt_name("A combobox field with 3 colors as options".to_string());
+    // Set the available options of this field, along with their display names
+    combobox_field.set_options(vec![
+        ChoiceOption::new("red".into(), Some("Red".into())),
+        ChoiceOption::new("green".into(), Some("Green".into())),
+        ChoiceOption::new("blue".into(), Some("Blue".into())),
+    ]);
+    // Allow users to add their own values instead.
+    combobox_field.set_edit(true);
```
reknih, line 70: Do combo-boxes need `DA` as well? How do they otherwise appear?

crates/krilla/src/surface.rs
reknih, line 179: Should refer to `Surface::start_variable_text` in its panic comment

crates/krilla/src/surface.rs
```diff
@@ -236,6 +238,23 @@ impl<'a> Surface<'a> {
...
+    /// Panics if a tagged/marked section has already been started (either via this function or via [`Surface::start_tagged`]).
+    pub fn start_variable_text(&mut self) {
+        self.bd.get_mut().start_marked_content(Name(b"Tx"));
+    }
+
+    /// End the current variable text marked section.
+    ///
+    /// # Panics
+    /// Panics if no variable text marked section has been started.
+    #[allow(missing_docs)]
```
reknih, line 253: Vestigial `allow(missing_docs)`?

crates/krilla/examples/forms.rs
reknih, line 1: A few compat issues:

**Acrobat 26.2.21931.0 on macOS:**

Image

https://github.com/user-attachments/assets/a4f7bd9d-3927-4bcb-9b64-709d2d7c0aa8

- Text appears upside down before any edits and uses wrong font. Dynamic appearance only used after first edit
- Checkbox and radio appearance stream is only used when checkbox is focused

**Preview 11.0 on macOS 27.0:**

Image

https://github.com/user-attachments/assets/c490c162-cb54-4772-a642-22176b9b861d

(Don't know what we can do about these)

- Field appearances rendered at low resolution (I fathom we cannot do anything there)
- Font setting in dynamic appearance ignored
- Text alignment ignored
- When writing in or after having changed text input, dynamic appearance unused, including when focus left
- When writing in a word text input, cursor set back to end after each keystroke
- When selecting from

**Chrome 153.0.8010.48:**

Text uses wrong font before edits. Dynamic appearance only used after first edit.

---

I think the issues with the appearance before edits stem from the fact that the `/AP` dictionaries' initial content stream differs in `Tf` arguments and contains a `Tm` operator not present in `DA`. Furthermore, when using `DA` to replace the contents of the `/Tx` marked content sequence, and following the interpretation of the example in PDF 2.0, 12.7.4.3, only the operators between `BT` and `ET` are replaced,  so we get a duplicate, if harmless, `rg` operator.

I see that you want to demonstrate (?) that variable appearance and the initial appearance stream can differ in their stylings, but when a reader uses DA vs the initial appearance (or its own) seems implementation dependant and should not be relied upon.

crates/krilla/src/interactive/form.rs
```diff
@@ -613,4 +1004,359 @@ pub mod kind {
...
+                                }
+                                Color::Regular(RegularColor::Luma(l)) => {
+                                    content.set_fill_gray(l.to_pdf_color());
+                                }
+                                Color::Regular(RegularColor::Cmyk(c)) => {
+                                    let comps = c.to_pdf_color();
+                                    content.set_fill_cmyk(comps[0], comps[1], comps[2], comps[3]);
+                                }
+                                Color::Special(_) => {
+                                    panic!("Device color space cannot be used with special colors")
```
reknih, line 1230: Shouldn't the error message be reversed?

crates/krilla/src/interactive/form.rs
```diff
@@ -613,4 +1004,359 @@ pub mod kind {
...
+            }
+
+            if let Some(paint) = &self.paint {
+                match &paint.0 {
+                    InnerPaint::Color(color) => {
+                        let color_space = color.color_space(sc);
+                        let color_space = sc.register_colorspace(chunk_container, color_space);
+
+                        match color_space {
+                            MaybeDeviceColorSpace::DeviceGray
```
reknih, line 1215: Are there opportunities for code reuse with other code elsewhere?

crates/krilla/src/interactive/form.rs
```diff
@@ -613,4 +1004,359 @@ pub mod kind {
...
+        resource::ResourceDictionaryBuilder,
+        serialize::{MaybeDeviceColorSpace, SerializeContext},
+        text::StandardFont,
+        util::NameExt,
+    };
+
+    #[derive(Debug, Clone, Default)]
+    pub(super) struct VariableText {
+        pub(super) appearance: VariableAppearance,
+        pub(super) appearance_buf: Option<Buf>,
```
reknih, line 1143: Should not serializing this be an action that takes `Self` and yields a new struct? Otherwise, changes to `appearance` after calling `serialize` may make `appearance_buf` stale

crates/krilla/src/interactive/form.rs
```diff
@@ -613,4 +1004,359 @@ pub mod kind {
...
+            &self,
+            sc: &mut SerializeContext,
+            chunk_container: &mut ChunkContainer,
+            rd_builder: &mut ResourceDictionaryBuilder,
+        ) -> Buf {
+            let identifier = match &self.font {
+                // TODO support fully embedding type 0 fonts
+                // FormFont::Embedded(font) => todo!(),
+                FormFont::Standard(standard_font) => {
+                    crate::text::FontIdentifier::Standard(*standard_font)
```
reknih, line 1188: Validation error for PDF/A?

crates/krilla/examples/forms.rs
diogotcorreia: Yes, but `VariableAppearance` has a default value of Helvetica, font size 0, so I omitted it in the example. I'll add an explicit call so that it isn't forgotten

Setting a font size other than 0 might fix some of the rendering issues in some viewers, so I'll try that too

crates/krilla/src/interactive/form.rs
diogotcorreia: Since this is internal to the crate I did not bother to do that, given the `FormField` struct is moved via `Document::set_field_tree`, so external callers can never mutate `appearance` after serialization.

But it might aid future refactors to have such a pattern, so I will add a new enum which ensures the appearance is always up-to-date

crates/krilla/src/interactive/form.rs
diogotcorreia: I believe it does not apply here, since we are not using the Type 1 fonts to render anything:
> The font programs for all fonts used for rendering within a conforming file shall be embedded within that file, as defined in ISO 32000-1:2008, 9.9. A font is considered to be used if at least one of its glyphs is referenced from a content stream (6.2.2).
> \- From PDF/A-3 Section 6.2.11.4.1

We are only referencing these fonts in the `/DA` field, which shall be ignored by PDF/A compliant readers:

> A conforming reader shall render the appearance dictionary without regard to any other keys and values in the annotation dictionary and shall ignore the values of the C, IC, Border, BS, BE, CA, H, DA, Q, DS, LE, LL, LLE, and Sy keys.
> \- From PDF/A-3 Section 6.3.3

I also checked the other PDF/A versions, here are the relevant sections:
- PDF/A-1: section 6.3.4; even though the `DA` field is not explicitly ignored, section 6.9 says
  > Every form field shall have an appearance dictionary associated with the field's data. A conforming reader shall render the field according to the appearance dictionary without regard to the form data.
- PDF/A-2: sections 6.2.11.4.1 and 6.3.3
- PDF/A-4: sections 6.2.10.4.1; similarly to PDF/A-1, the `DA` field is not explicitly ignored, but section 6.3.3 refers to "ISO 32000-2:—, 12.5.2", which says:
  > A PDF reader shall render the appearance dictionary without regard to any other keys and values in the annotation dictionary and shall ignore the values of the C, IC, Border, BS, BE, BM, CA, ca, H, DA, Q, DS, LE, LL, LLE, and Sy keys.

What do you think?

TL;DR, given this font is only used in `/DA` and not in a content stream, I believe it can be included in a PDF/A conformant file.

crates/krilla/src/interactive/form.rs
diogotcorreia: Yes, this code is a modified copy of `crates/krilla/src/content.rs` lines 1168-1189, but that function is inside another one, so it can't be accessed easily.

I'll try to refactor, but not sure where to put it 😅 And the `ContentColorSpace` enum it takes is also currently private to that file.

crates/krilla/src/interactive/form.rs
diogotcorreia: I agree, but it was copied from existing code as said above

crates/krilla/examples/forms.rs
diogotcorreia: > **Acrobat 26.2.21931.0 on macOS:**
> - Text appears upside down before any edits and uses wrong font. Dynamic appearance only used after first edit

Seems like Acrobat is not respecting the `/Matrix` field in the XObject. I will try to check for alternatives, but it might mean we will have to break krilla's inverted y-axis abstraction. Afaict this only impacts text fields, all other field types work fine with `/Matrix`.

Yes, the `/DA` field is only used when writing a new value. I had a bug in the example where a sans-serif font was used for the appearance stream while a serif font (Times New Roman) was used in `/DA`, but that is now fixed. While the serif font we are using is not metric compatible with Times New Roman afaict, I did not want to add a new font to the repo. If you think the example would benefit from it, I can add and use, e.g., Liberation Serif.

> - Checkbox and radio appearance stream is only used when checkbox is focused

This is Acrobat's behavior afaik, it overlays its own appearance when unfocused. It works fine for me when printing/saving the file.

> **Preview 11.0 on macOS 27.0:**
> - Field appearances rendered at low resolution (I fathom we cannot do anything there)
> - Font setting in dynamic appearance ignored
> - Text alignment ignored
> - When writing in or after having changed text input, dynamic appearance unused, including when focus left
> - When writing in a word text input, cursor set back to end after each keystroke

Sadly I don't know what we can do here. Have you checked if at least the styles are better when printing? What happens if you edit and then open with a separate editor?

> - When selecting from

I think your sentence was cut off


> **Chrome 153.0.8010.48:**
> - Text uses wrong font before edits. Dynamic appearance only used after first edit.

This was/is expected, same as Acrobat

diogotcorreia (commented):
I left some of the changes as fixup commits without rebasing, so you can more easily diff. I'll rebase/force-push when the PR is approved.

Problems remaining:
- Acrobat not respecting `/Matrix` in text fields; I'll investigate further what the best solution is
- Code reuse for the color serialization; suggestions welcome

I hope I have addressed everything else

crates/krilla/src/interactive/form.rs
reknih: Is veraPDF happy with this under all versions? If so, I'm fine with it

crates/krilla/examples/forms.rs
reknih: >> When selecting from
>
> I think your sentence was cut off

This was about the list box I think where, when selecting, the choices disappear altogether

diogotcorreia: > This was about the list box I think where, when selecting, the choices disappear altogether

I think this is solved by setting a font size too, can re-test now that I've fixed the Acrobat issues too?

crates/krilla/src/interactive/form.rs
diogotcorreia: Yes, veraPDF is happy with this in all PDF/A versions!

We should handle PDF 2.0 however, I believe we would need some more font info for PDF 2.0 (`/FirstChar`, `/LastChar`, `/Widths`). As with RADIOS_IN_UNISON in the last PR, I don't think we have the tooling to throw validation errors on specific PDF versions. Should I just leave a TODO, or panic?
Or perhaps it might be time to add proper tooling

diogotcorreia (commented):
I decided to force push since I think it would have been more confusing without.

I addressed the comments, but I also changed my refactor of `ResourceDictionaryBuilder` to add the prefix to the `ResourceDictionary` itself instead of each of the Resources, which simplifies the code a lot and also allows for easier reuse of existing code elsewhere (e.g., the fill color serialization from `content.rs`)

crates/krilla/src/interactive/form.rs
diogotcorreia: I re-used the code from `content.rs` in 38c47fa, let me know if you prefer me to revert.

crates/krilla/src/stream.rs
```diff
@@ -112,6 +112,21 @@ impl<'a> StreamBuilder<'a> {
...
+
+    /// Get the surface of the stream builder.
+    /// Unlike [`StreamBuilder::surface`], the returned surface has its y-axis inverted similarly to
+    /// a [page surface](crate::page::Page::surface).
+    pub fn surface_with_bbox(&mut self, bbox: Rect) -> Surface<'_> {
+        let transform = Transform::from_row(1.0, 0.0, 0.0, -1.0, 0.0, bbox.height());
+        let mut content_builder = ContentBuilder::new(transform, false, self.sc);
+        content_builder.expand_bbox(bbox);
+        self.surface_with_content_builder(content_builder)
+    }
```
diogotcorreia, line 127: @LaurenzV What is on your opinion on this function? More concretely:

- Is it too weird that this has the y-axis inverted while `StreamBuilder::surface` does not?
- Do you prefer this to be here, or in `Surface` instead (e.g., as `Surface::stream_builder_with_bbox`)?

diogotcorreia:
Regarding pdfcpu (https://github.com/LaurenzV/krilla/pull/430#issuecomment-5701961816), the issue I submitted seems to be resolved.
However, upgrading to the just released pdfcpu 0.16.0-rc.1 seems to raise a lot of validation errors regarding dates:
```
validating(mode=strict) snapshots/pdf_version_14_tagged.pdf ...
validate snapshots/pdf_version_14_tagged.pdf: validation error (obj#:14) (try --mode=relaxed): metadata/info order: date object: <D:20010101000000Z> invalid date
```

I'll investigate and see if it's our problem or if I should submit another issue upstream.

EDIT: nvm, we have our own fixes for this, we just need to rebase them when updating the fork: https://github.com/LaurenzV/pdfcpu/commit/59b622105a4d9baa35bd4eaf07c0d8650ffecc1f

reknih:
Screenshot 2026-09-23 at 15 00 18

In Acrobat, text now appears right-side up before editing. The form is functional. Fields without dynamic appearances use our custom appearance streams only when in focus or when printed.

Screenshot 2026-09-23 at 15 03 27

Save for the font choice, pdf.js behaves correctly.

Screenshot 2026-09-23 at 15 04 31

Pdfium behaves correctly.

Screenshot 2026-09-23 at 15 06 15

Quartz still exhibits all previously observed faults, including the inability to select from the list box.

crates/krilla/examples/forms.rs
reknih: retested, see below, I think Quartz is just a piece of shit for forms.

crates/krilla/src/interactive/form.rs
reknih: > Should I just leave a TODO, or panic? Or perhaps it might be time to add proper tooling

The way I know @LaurenzV, he would like that to happen in another PR. In the meantime, panic instead of writing invalid files.

reknih: > Yes, veraPDF is happy with this in all PDF/A versions!

Then we are good to go in that respect

diogotcorreia:
> Save for the font choice, pdf.js behaves correctly.

From my testing, the font is used correctly when printing, it's just the user-facing UI that uses a different font.

LaurenzV:
Updated pdfcpu.

